PR #10, opened 2020-01-11
Fix #6 Handle 'eval/compile' by accessing the outer frame

### LINKED ISSUE #6: crashes on eval/compile
There are some problems with code which use eval/compile:
``` python
import attr

while True:
    @attr.s
    class T:
        x = attr.ib(default=666)
```
```
$ python -m scalene issue.py
```
```
File "/home/poh/.pyenv/versions/3.7.2/lib/python3.7/site-packages/attr/_make.py", line 245, in _make_attr_tuple_class
    eval(compile("\n".join(attr_class_template), "", "exec"), globs)
  File "", line 1, in <module>
  File "/home/poh/.pyenv/versions/3.7.2/lib/python3.7/site-packages/scalene/scalene.py", line 91, in cpu_signal_handler
    if not scalene.should_trace(fname):
  File "/home/poh/.pyenv/versions/3.7.2/lib/python3.7/site-packages/scalene/scalene.py", line 129, in should_trace
    if filename[0] == '<':
IndexError: string index out of range
```
[Attrs](https://github.com/python-attrs/attrs) is very famous library. It is «must have» to support it.

friday-james:
'eval/compile' gives no `f_code.co_filename`. We have to look back into the outer frame in order to check the `co_filename`. Hence a line is added to look back by one frame if the `co_filename` is empty. This fixes #6 .

### DIFF
--- a/scalene/scalene.py
+++ b/scalene/scalene.py
@@ -61,6 +61,7 @@ class scalene(object):
     total_cpu_samples      = 0           # how many CPU    samples have been collected.
     total_malloc_samples   = 0           # "   "    malloc "       "    "    "
     total_free_samples     = 0           # "   "    free   "       "    "    "
+    current_footprint = 0
     signal_interval        = 0.01        # seconds between interrupts for CPU sampling.
     elapsed_time           = 0           # total time spent in program being profiled.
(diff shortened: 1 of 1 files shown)


emeryberger:
Thanks!

PR #21, opened 2020-01-16
Memoize should_trace for faster execution

groutr:
This PR caches the output of `should_trace` which yields a slight performance increase.
Running the julia1_nopil.py benchmark on my laptop:
master: 8.14s
memoize_shouldtrace: 7.98s

Running the benchmark and looking a the cache statistics shows that the cache is hit 530  times for 1 file.

`should_trace` can perform two string searches before returning a result. `os.path.abspath` isn't exactly the speediest function either.  Caching the results of `should_trace` makes a lot of sense.

I used the LRU cache available in the standard library. I feel the default maxsize of 128 is sufficient.


emeryberger:
Good catch. Thanks!

PR #22, opened 2020-01-17
Output handling improvements

groutr:
Slightly modifies the output handling of scalene to:
1. Avoid reading the entire source file into memory, instead processing it line by line.
2. Use enumerate to track the line number.
3. Use str.rstrip to more reliably remove all trailing whitespace.
4. Fixes a possible divide by zero.  If `this_cpu_samples` is zero then the expression for `percent_cpu_time` will evaluate to zero anyway. Further down, there is a check for `Scalene.total_cpu_samples` being zero.  I'm guessing what was meant to be checked here was `Scalene.total_cpu_samples`.


emeryberger:
Many thanks!

PR #23, opened 2020-01-17
Fix usage example for linux.

groutr:
Usage example for linux had a .dylib extension instead of .so

Used dedent to have nicer formatting of the usage string.


emeryberger:
Thanks!

PR #41, opened 2020-04-09
add docs/README_CN.md

Insutanto:
I have translated the README from English to Chinese.
It may be convenient for Chinese developers to know about scalene.


Insutanto:
> I'd love to include this - can you commit to keeping the README updated? I am planning on making some changes to the current README, and I'd like to see those reflected in this doc if I add it to the repo.

Yes, of course. scalene is very convenient and I am going to use it in my work. So, I can keep the README updated.

PR #44, opened 2020-04-19
Print output time and sample counts in header

Insutanto:
To know the output time and sample counts. Sometime, I can't know the scalene has sampled my program or not.


emeryberger:
The latest version should always indicate when Scalene has run. Please pull and verify.

PR #59, opened 2020-06-09
Add install_requires

Insutanto:
`setup.py` file misses the `install_requires` which will cause a `ModuleNotFoundError` when the `rich` package not in the user environment.

--------------------------------------------------
```
    from rich.console import Console
ModuleNotFoundError: No module named 'rich'
```


emeryberger:
Thanks!

PR #75, opened 2020-09-23
expose errors when calling main

boegel:
Problems that occur when calling `scalene.Scalene.main` should not be swept under the rug, but should be exposed as an error and trigger a non-zero exit code.

I'm not 100% sure the proposed change is OK, but the current code was very effective at hiding a problem with my installation of `scalene` 1.1.1 (see #76).

### DIFF
--- a/scalene/__main__.py
+++ b/scalene/__main__.py
@@ -1,9 +1,12 @@
+import sys
+
 def main():
     try:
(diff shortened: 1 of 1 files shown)


emeryberger:
LGTM - thanks!

PR #80, opened 2020-09-29
Adding error message in case subprocess receives a signal

jaltmayerpizzorno:
I've been debugging a case where libscalene.so gets a SIGSEGV. In this case, no error message is printed, but the main process just exits -11.  I suggest these changes to clarify what is going on.


emeryberger:
Thanks! Let me know what you find with respect to the seg fault.

PR #88, opened 2020-11-06
Fix naming conflicts and add in poll shim

sternj:
Renamed `scalene.py` to `entrypoint.py` to prevent some conflicts observed in importing with benchmarks. Made benchmark import relative. Added in `ReplacementPollSelector` to avoid dropped signals in long-running `poll`s.

### DIFF
--- a/scalene/__main__.py
+++ b/scalene/__main__.py
@@ -1,12 +1,14 @@
 import sys
 
+
 def main():
     try:
-        from scalene import scalene
-        scalene.Scalene.main()
+        from scalene import entrypoint
+        entrypoint.Scalene.main()
     except Exception as exc:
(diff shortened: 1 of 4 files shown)


scalene/__main__.py
```diff
@@ -1,12 +1,14 @@
 import sys
 
+
 def main():
     try:
-        from scalene import scalene
-        scalene.Scalene.main()
+        from scalene import entrypoint
```
emeryberger, line 6: Maybe would be better to call it `profiler`? (`entrypoint` is pretty nondescript.)

emeryberger: Or even `scalene_profiler`...

emeryberger (changes requested):

emeryberger (approved):

PR #93, opened 2020-11-24
Propagate exceptions

sternj:
Proposing that this check be removed, as many programs propagate errors to the client with unhandled exceptions


emeryberger:
So this is printing a traceback; is this the right default behavior?

PR #95, opened 2020-12-08
Add support for `fork` context of multiprocessing library

sternj:
- Add support for `fork` context of built-in multiprocessing library by adding support for a new signal, `SIGTSTP`, which signifies that a `fork` system call has been executed and the current process is a child. 
- refines temporary file removal so that cleanup only occurs in root Scalene process
- Refines thread enumeration to prevent deadlock after fork, at the cost of perhaps dropping some signals. Needs further work.
- Adds in more shims to support new signal and exit modes

### DIFF
--- a/scalene/replacement_fork.py
+++ b/scalene/replacement_fork.py
@@ -13,6 +13,6 @@ def replacement_fork(scalene: Scalene):
     def fork_replacement():
         result = orig_fork()
         if result == 0:
-            signal.raise_signal(signal.SIGTSTP)
+            signal.raise_signal(Scalene.fork_signal)
         return result
     os.fork = fork_replacement
--- a/scalene/replacement_fork.py
+++ b/scalene/replacement_fork.py
@@ -4,6 +4,10 @@ import signal
 
 @Scalene.shim
 def replacement_fork(scalene: Scalene):
+    """
+    Raises a signal when a process is the child after
+    a fork system call.
+    """
     orig_fork = os.fork
 
     def fork_replacement():
--- a/scalene/replacement_fork.py
+++ b/scalene/replacement_fork.py
@@ -1,0 +1,14 @@
+import os
+from scalene.scalene_profiler import Scalene
+import signal
+
+@Scalene.shim
+def replacement_fork(scalene: Scalene):
+    orig_fork = os.fork
+
+    def fork_replacement():
+        result = orig_fork()
+        if result == 0:
+            signal.raise_signal(signal.SIGTSTP)
+        return result
+    os.fork = fork_replacement
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -186,6 +186,13 @@ def parse_args() -> Tuple[argparse.Namespace, List[str]]:
         default=100,
         help="only report profiles with at least this many allocations (default: 100)",
     )
+    parser.add_argument(
+        "--malloc-replacement",
+        dest="malloc_replacement",
+        type=str,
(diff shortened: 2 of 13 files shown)


scalene/scalene_profiler.py
```diff
@@ -224,6 +231,8 @@ def parse_args() -> Tuple[argparse.Namespace, List[str]]:
             os.environ["LD_PRELOAD"] = os.path.join(
                 os.path.dirname(__file__), "libscalene.so"
             )
+            if arguments.malloc_replacement:
+                os.environ["LD_PRELOAD"] += ":" + arguments.malloc_replacement
```
emeryberger: Need to also do this for OS X.

emeryberger (changes requested):

scalene/replacement_fork.py
```diff
@@ -0,0 +1,14 @@
...
+import signal
+
+@Scalene.shim
+def replacement_fork(scalene: Scalene):
+    orig_fork = os.fork
+
+    def fork_replacement():
+        result = orig_fork()
+        if result == 0:
+            signal.raise_signal(signal.SIGTSTP)
```
emeryberger, line 16: Annoying to have the signal (SIGTSTP) here and in the other file - would be better if it were in only one place.

scalene/scalene_profiler.py
```diff
@@ -408,9 +411,9 @@ def get_original_lock() -> threading.Lock:
     # path for the program being profiled
     __program_path: str = ""
     # temporary directory to hold aliases to Python
-    __python_alias_dir: Any = tempfile.TemporaryDirectory(prefix="scalene")
+    __python_alias_dir: Any = tempfile.mkdtemp(prefix="scalene")
```
emeryberger, line 405: Can't recall - is this still in use?

scalene/scalene_profiler.py
```diff
@@ -829,6 +837,8 @@ def compute_frames_to_record(
         this_frame: FrameType,
     ) -> List[Tuple[FrameType, int, FrameType]]:
         """Collects all stack frames that Scalene actually processes."""
+        if threading._active_limbo_lock.locked():
```
emeryberger, line 831: Needs an exploratory comment.

scalene/scalene_profiler.py
```diff
@@ -862,6 +872,7 @@ def compute_frames_to_record(
                 back = cast(FrameType, frame.f_back)
                 fname = Filename(back.f_code.co_filename)
             while not Scalene.should_trace(fname):
+
```
emeryberger, line 866: Omit space  — use " make black" before commit.

emeryberger (changes requested):

scalene/scalene_profiler.py
sternj: this was removed in another commit, it was from my attempt to add in a replacement memory allocator

scalene/scalene_profiler.py
sternj: [yes](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp)

scalene/scalene_profiler.py
emeryberger: gotcha

sternj:
Note-- as of `fb79800`, not ready to merge due to race condition in allocator signal writing

sternj:
Closing this so I can start on a clean branch, since attempting to rebase to #102 and #103 polluted the tree

PR #100, opened 2020-12-20
Make memcpy nondeterministic

sternj:


include/memcpysampler.hpp
```diff
@@ -108,6 +108,7 @@ class MemcpySampler {
     if (unlikely(sampleMemop)) {
       writeCount();
       _memcpyTriggered++;
+      _memcpyOps = 0;
```
emeryberger, line 111: Do we need this anymore? I think it's obsoleted by using the sampler.

sternj: memcpyOps is part of the payload written to the file, and seems to be used by the profiler

emeryberger: Just read it more carefully and you are correct! Thanks.

PR #102, opened 2020-12-21
Moved memcpy signal file and counter to mmap, abstracted out writing to mmaped files

sternj:
- created SampleFile class
- made SampleHeap and MemCpySampler use SampleFile for logging
- Made scalene_profiler read from memcpy log in `mmap`ed region as opposed to directly from file


include/memcpysampler.hpp
```diff
@@ -45,9 +39,9 @@ class MemcpySampler {
     return len;
   }
   
-  ~MemcpySampler() {
-    unlink(scalene_memcpy_signal_filename);
-  }
+   ~MemcpySampler() {
+   _samplefile.~SampleFile();
```
emeryberger, line 43: Why do you need to explicitly call the destructor?

emeryberger: (You don't: https://isocpp.org/wiki/faq/dtors#calling-member-dtors)

include/samplefile.hpp
```diff
@@ -0,0 +1,71 @@
...
+    public:
+        SampleFile(char* filename_template, char* lockfilename_template) {
+            auto pid = getpid();
+            stprintf::stprintf(_signalfile, filename_template, pid);
+            stprintf::stprintf(_lockfile, lockfilename_template, pid);
+            _fd = open(_signalfile, flags, perms);
+            _lock_fd = open(_lockfile, flags, perms);
+            ftruncate(_fd, MAX_FILE_SIZE);
+            _mmap = reinterpret_cast<char*>(mmap(0, MAX_FILE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, _fd, 0));
+            _lastpos = reinterpret_cast<int*>(mmap(0, MAX_FILE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, _lock_fd, 0));
```
emeryberger, line 32: `_lastpos` is just supposed to hold an `int` (which honestly should be a `uint64_t` or something for clarity), so you don't need to allocate `MAX_FILE_SIZE` bytes for it -- just `sizeof(_lastpos)`.

include/samplefile.hpp
```diff
@@ -0,0 +1,71 @@
...
+                abort();
+            }
+        }
+        ~SampleFile() {
+            unlink(_signalfile);
+            unlink(_lockfile);
+        }
+        void writeToFile(char* line) {
+            lock.lock();
+            rte_memcpy(_mmap + *_lastpos,
```
emeryberger, line 49: I have `rte_memcpy` disabled for Darwin builds so you'll need the same logic here (or just write the obvious loop instead -- probably preferable).

include/samplefile.hpp
```diff
@@ -0,0 +1,71 @@
...
+                MAX_BUFSIZE);
+            *_lastpos += strlen(_mmap + *_lastpos) - 1;
+            lock.unlock();
+        }
+
+    private:
+        static constexpr auto flags = O_RDWR | O_CREAT;
+        static constexpr auto perms = S_IRUSR | S_IWUSR;
+
+        char _signalfile[256];
```
emeryberger, line 60: Comments for all members please :).

include/sampleheap.hpp
```diff
@@ -49,22 +53,12 @@ class SampleHeap : public SuperHeap {
...
-      tprintf::tprintf("Scalene: internal error = @\n", errno);
-      abort();
-    }
+
   }
 
   ~SampleHeap() {
     // Delete the log file.
-    unlink(scalene_malloc_signal_filename);
+    _samplefile.~SampleFile();
```
emeryberger, line 61: Not needed.

scalene/scalene_profiler.py
```diff
@@ -880,7 +883,9 @@ def malloc_signal_handler(
         this_frame: FrameType,
     ) -> None:
         """Handle malloc events."""
+        # print("Locking malloc")
```
emeryberger, line 886: Please replace these commented out print statements with comments, if appropriate.

emeryberger (changes requested):

include/samplefile.hpp
emeryberger: Bump @sternj

sternj: Acknowledged, and getting it done tomorrow!

include/samplefile.hpp
```diff
@@ -46,9 +46,11 @@ class SampleFile {
         }
         void writeToFile(char* line) {
             lock.lock();
-            rte_memcpy(_mmap + *_lastpos,
-                line,
-                MAX_BUFSIZE);
+            char* src = line;
+            char* dst = _mmap + *_lastpos;
+            for(int i = 0; i < MAX_BUFSIZE; i++) {
```
emeryberger, line 51: Why is so much data being copied? I would have thought it would just be copying the length of the string.

emeryberger (changes requested):

include/samplefile.hpp
sternj: This is the parameterization used in your original code! There's garbage at the end of the signal files that I think is the result of this.

emeryberger: Ha! Awesome. Might as well not perpetuate the bug I guess :).

emeryberger: OK, will merge and fix later.

emeryberger:
Thanks for the contributions! I had to revert the merge to figure out some issues, as it wasn't working on all platforms. Among updates:
* Changed lock file from `MAP_ANONYMOUS` (which doesn't use an fd), added a needed `ftruncate`, changed size of the lock file to 4k
* Backed out use of `rte_memcpy` in favor of MUSL's pure C `memcpy`, lightly ported to C++ :) 
* Changed to use `strncpy` instead of copying all bytes, which should make updates a lot faster for samples (though the `n` argument is not quite right and needs to be fixed, marked with a `FIXME`)
* Added support for Linux + ARM targets
* Changed the RNG to not use the system clock stuff, which was leading to a segfault (back to using `getpid` and `pthread_self` as the seeds, plus a big ol' number)

PR #103, opened 2020-12-24
Revert "Moved memcpy signal file and counter to mmap, abstracted out writing to mmaped files"

emeryberger:
Reverts emeryberger/scalene#102

Getting failures on some tests (test/original/bm_spectral_norm.py).


emeryberger:
...and not fixed by reverting...

PR #105, opened 2020-12-28
Synchronize with shared spinlock

sternj:
Added support for spinlock shared in `mmap` region as synchronization primitive, then implemented it to see impact on performance of synchronizing all reads and writes to each signal file.


sternj:
@emeryberger tested and executed on x86 mac, ready to merge

sternj:
Closing because of issues with rebase

PR #108, opened 2021-01-02
Add explanation for result columns

madig:
Closes https://github.com/emeryberger/scalene/issues/107.

What do the % mean in the "Memory usage over time / %" column?


emeryberger:
Thanks!

PR #118, opened 2021-02-04
Add support for fork context of multiprocessing

sternj:
Makes performance from child processes correctly reported


emeryberger:
Let me know when this is safe to merge!

sternj:
@emeryberger good to go!

PR #127, opened 2021-02-16
Removed race conditions and double initialization

sternj:


emeryberger:
Great work - thanks!!!

PR #137, opened 2021-03-05
Changes NextHeap initialization to compensate for "early" malloc invocation

jaltmayerpizzorno:
Malloc can be used very early during process startup, even before the C++ constructors run.
This was causing NextHeap to initialize twice, and StaticBufferHeap to fail if actually needed.
This code adjusts for that, mostly by relying on static data members.

As discussed, this PR also renames NextHeap to SysMallocHeap, which prevents the diff from working.
The following link shows most of the actual changes: https://github.com/plasma-umass/scalene/commit/2f351dd9f4bba75720da96227b295baf229ffce2

I think we should remove the __APPLE__ variant, but I would like to improve my testing before opening a PR.


jaltmayerpizzorno:
Mm, I think the malloc wrappers need to do that initialization spiel, not SysMallocHeap/NextHeap. Let me take a look at that first...

PR #138, opened 2021-03-05
Fixed some issues caused by fetching line numbers

sternj:


scalene/scalene_profiler.py
```diff
@@ -877,6 +877,7 @@ def allocation_signal_handler(
 
         for (frame, _tident, _orig_frame) in new_frames:
             fname = Filename(frame.f_code.co_filename)
+            print(f"orig name: {fname}")
```
emeryberger, line 880: Lots of debug print statements that need to be squashed

emeryberger (changes requested):

PR #141, opened 2021-03-10
Added GPU profiling support.

emeryberger:
Using GPUtil to extract load; currently set to do this at 1/10th the CPU sampling rate to reduce overhead. Also fixed zero check for system time.


emeryberger:
A note: this support is currently limited to nVidia.

PR #147, opened 2021-03-15
Revamped heap redirection code

jaltmayerpizzorno:
This PR:

- adds a thread-local boolean to detect when `malloc`/etc. is called recursively from a custom heap's constructor or `malloc()`;
- redirects such recursive calls to `StaticBufferHeap`;
- adds mutex to `StaticBufferHeap`, just in case recursive mallocs happen from multiple threads;
- simplifies `NextHeap`, given the changes above, and renames it `SysMallocHeap`;
- documents the rationale;

This should probably be merged into [Heap-Layers](https://github.com/plasma-umass/Heap-Layers) eventually.

I tested it manually on Linux and MacOS; we really need automated tests for these things -- maybe after that merge into Heap-Layers?


include/staticbufferheap.hpp
```diff
@@ -34,6 +41,10 @@ class StaticBufferHeap {
     return ptr;
   }
 
+  void *memalign(size_t alignment, size_t sz) {
+    return malloc(sz); // FIXME 'alignment' ignored
```
emeryberger, line 45: Since we are not implementing `free` (that is, it's a NOP), then we can actually implement `memalign` here just in case by allocating a sufficiently-large chunk of memory and then placing the allocated object in the right spot within it. Anyway, this should arguably be using a lock layer around https://github.com/emeryberger/Heap-Layers/blob/master/heaps/top/staticheap.h, right?

include/sysmallocheap.hpp
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+#ifndef NEXTHEAP_HPP
```
emeryberger, line 3: should be renamed to `SYSMALLOCHEAP_HPP`

include/heapredirect.hpp
```diff
@@ -0,0 +1,123 @@
+#pragma once
+
+#ifndef HEAPREDIRECT_H
+#define HEAPREDIRECT_H 
+
+#include "staticmutex.hpp"
+#include "staticbufferheap.hpp"
+
```
emeryberger, line 8: It pains me for this whole thing to be a giant macro... if more of the code could be factored out and only a tiny bit actually be in a macro, that would be better.

emeryberger (changes requested):

include/staticbufferheap.hpp
jaltmayerpizzorno: Good point about just fixing`memalign`.

We don't need a separate lock guard here as long as we invoke `malloc`, since that method has it.

include/sysmallocheap.hpp
jaltmayerpizzorno: Oops

include/staticbufferheap.hpp
emeryberger: Right, I meant that the whole thing could just be a `LockedHeap` around a slightly modified `StaticHeap`.

include/heapredirect.hpp
jaltmayerpizzorno: Sure, can I move some of that to a `.cpp` or would you prefer a `class` template?  I had it in a class template before, but that had some downsides (most easily remedied with C++17's `inline static` once again).

We do have to emit the `xxmalloc`, `xxfree`, etc. Because the various `Heap`s lack a common interface (base and `virtual` methods), we can't use a factory method for the heap and move them to a `.cpp`... it's easier to just leave them in the macro, even if all they do is invoke class template methods.

Let me know and I'll be happy to adjust.

emeryberger: Yeah, I think a macro to emit the `xxmalloc` things is fine; class template also best IMHO (to the extent that this can be header-only (mostly), better).

include/staticbufferheap.hpp
jaltmayerpizzorno: Ah yes. I'll check whether `std::mutex` and `std::lock_guard` are ok to use... after all the difficulties with `std::call_once` and `thread_local` I went all `pthread`. If that works, we'd just need to modify it for alignment (in `malloc` and in `memalign`).

emeryberger: If you use `LockedHeap`, you can just use any class with the same interface as `std::mutex`: see https://github.com/emeryberger/Heap-Layers/blob/master/heaps/threads/lockedheap.h.

jaltmayerpizzorno: > Since we are not implementing `free` (that is, it's a NOP), then we can actually implement `memalign` here just in case by allocating a sufficiently-large chunk of memory and then placing the allocated object in the right spot within it.

Actually, not exactly, as `getSize` wouldn't work, but we can skip enough in the buffer to arrange for the alignment. Working on it.

jaltmayerpizzorno: > If you use `LockedHeap`, you can just use any class with the same interface as `std::mutex`: see https://github.com/emeryberger/Heap-Layers/blob/master/heaps/threads/lockedheap.h.

Could you modify `LockedHeap` to also wrap-lock around `memalign`?  Then I could refactor the locking out of `StaticBufferHeap`.

jaltmayerpizzorno: > Could you modify `LockedHeap` to also wrap-lock around `memalign`? Then I could refactor the locking out of `StaticBufferHeap`.

I've opened a PR for that on Heap-Layers.

emeryberger (approved):

emeryberger:
Great work!

include/staticbufferheap.hpp
emeryberger: Yep, merged!

PR #164, opened 2021-03-26
Fixed smoketest

sternj:


include/memcpysampler.hpp
```diff
@@ -184,7 +189,7 @@ class MemcpySampler {
         _memcpyTriggered(0) {
           static HL::PosixLock init_lock;
           init_lock.lock();
-    sighandler_t old_sig = signal(MemcpySignal, SIG_IGN);
+    SIGHANDLER_TYPE old_sig = signal(MemcpySignal, SIG_IGN);
```
emeryberger, line 192: why not just use `auto`?

heaplayers-make.mk
```diff
@@ -2,16 +2,21 @@
 CPPFLAGS = -std=c++14 -g -O3 -DNDEBUG -fno-builtin-malloc -fvisibility=hidden
 CXX = clang++
 
+UNAME_S := $(shell uname -s)
+UNAME_P := $(shell uname -p)
+ifeq ($(UNAME_S),Darwin)
+	ifeq ($(UNAME_P),arm64)
+		ARMFLAG = -arch arm64
```
emeryberger, line 9: great idea to do this!

include/sampleheap.hpp
```diff
@@ -54,11 +61,11 @@ class SampleHeap : public SuperHeap {
         _freedLastMallocTrigger(false) {
       
       signal_init_lock.lock();
-      sighandler_t old_malloc = signal(MallocSignal, SIG_IGN);
+      SIGHANDLER_TYPE old_malloc = signal(MallocSignal, SIG_IGN);
```
emeryberger, line 64: same comment as above, just use `auto` (or `decltype` if you really need a type declaration)

include/sampleheap.hpp
```diff
@@ -54,11 +61,11 @@ class SampleHeap : public SuperHeap {
...
       if (old_malloc != SIG_DFL) {
         signal(MallocSignal, old_malloc);
       }
-      sighandler_t old_free = signal(FreeSignal, SIG_IGN);
+      SIGHANDLER_TYPE old_free = signal(FreeSignal, SIG_IGN);
```
emeryberger, line 68: ditto

emeryberger (changes requested):

include/memcpysampler.hpp
sternj: In all honesty, I just forgot that auto existed. I'll switch it to that!

PR #173, opened 2021-04-14
Added check for top-level frame

sternj:


scalene/scalene_profiler.py
```diff
@@ -736,7 +736,7 @@ def enter_function_meta(
         f = frame
         while "<" in Filename(f.f_code.co_name):
             f = cast(FrameType, frame.f_back)
-            if "<genexpr>" in f.f_code.co_name:
+            if "<genexpr>" in f.f_code.co_name or "<modlue>" in f.f_code.co_name:
```
emeryberger, line 739: looks like a typo? ("modlue" instead of "module")

PR #179, opened 2021-05-04
Use Hoard as the allocator on MacOS.

jaltmayerpizzorno:
**This PR is intended for discussion primarily. This hasn't been extensively tested, Sam is also working on the same thing, his version may be better, or have elements we could incorporate here, or we could find out we don't need these changes.**


jaltmayerpizzorno:
We discussed it and tested it, so I guess we could merge this in now.

Issue #183, opened 2021-05-12
macOS 11.3: brew install: No rule to make target `vendor/Heap-Layers/wrappers/macwrapper.cpp'


andreas-zeller:
**Describe the bug**
`brew install scalene` fails on my machine

**To Reproduce**
```sh
$ brew tap plasma-umass/scalene
$ brew install --head plasma-umass/scalene/libscalene
==> Installing libscalene from plasma-umass/scalene
==> Cloning https://github.com/plasma-umass/scalene.git
Cloning into '/Users/zeller/Library/Caches/Homebrew/libscalene--git'...
==> Checking out branch master
Already on 'master'
Your branch is up to date with 'origin/master'.
==> git clone https://github.com/emeryberger/Heap-Layers
==> make
Last 15 lines from /Users/zeller/Library/Logs/Homebrew/libscalene/02.make:
2021-05-12 12:56:20 +0200

make

mkdir -p vendor && cd vendor && /usr/bin/true
mkdir -p vendor && cd vendor && git clone https://github.com/emeryberger/Hoard
mkdir -p vendor && cd vendor && git clone https://github.com/mpaland/printf
make: *** No rule to make target `vendor/Heap-Layers/wrappers/macwrapper.cpp', needed by `all'.  Stop.
make: *** Waiting for unfinished jobs....
Cloning into 'printf'...
Cloning into 'Hoard'...
cd vendor/Hoard/src && ln -s ../../Heap-Layers  # avoid inconsistencies by using same package

If reporting this issue please do so at (not Homebrew/brew or Homebrew/core):
  https://github.com/plasma-umass/homebrew-scalene/issues

Please create pull requests instead of asking for help on Homebrew's GitHub,
Twitter or any other official channels.
```

**Expected behavior**
`scalene` installs.

**Desktop (please complete the following information):**
 - OS: macOS 11.3.1

**Additional context**

jaltmayerpizzorno:
'Tag Andreas, we can reproduce the problem. It should be fixed shortly.

emeryberger:
Note that the instructions have changed very slightly (no more "lib").

    % brew install --HEAD plasma-umass/scalene/scalene

emeryberger:
Many thanks, @andreas-zeller, for the bug report!

emeryberger:
Note: https://github.com/plasma-umass/scalene/issues/747

PR #184, opened 2021-05-12
Fixed missing dependency causing 'brew install' to fail

jaltmayerpizzorno:


emeryberger:
Fixes https://github.com/plasma-umass/scalene/issues/183.

PR #191, opened 2021-05-18
Moves allocation and memcpy signal handling from starting threads to producer-consumer queue

jaltmayerpizzorno:
Note that this changes the minimum Python requirement to 3.7, as SimpleQueue is only available from that version onwards.
Alternatively, this could be modified to use my [signal_queue](https://github.com/jaltmayerpizzorno/signal_queue) implementation.

This also speeds up scalene a bit.

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -33,6 +33,7 @@ import tempfile
 import threading
 import time
 import traceback
+import queue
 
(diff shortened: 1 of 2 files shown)


jaltmayerpizzorno:
This needs more work to support the `multiprocessing` module. Closing for now so there's no confusion.

PR #194, opened 2021-05-19
Bugfix/cautious attribute retrieval (closes #193)

### LINKED ISSUE #193: Crashes if a profiled function defines a non-class using the name `cls`
**Describe the bug**
Scalene tries to retrieve `__name__` from any value named `cls`, even if this attribute is not present, causing an unhandled `AttributeError`

**To Reproduce**

Running Scalene on the following program:

```python
# repro.py
import time

def f():
    cls = "This value isn't a class"
    time.sleep(0.5)  # to be sure this frame gets sampled

f()
```

yields the output:

```
$ python3 -m scalene repro.py 
 'str' object has no attribute '__name__'
  File "/home/snoopjedi/repos/scalene/scalene/scalene_profiler.py", line 1241, in profile_code
  File "repro.py", line 8, in <module>
    f()
  File "repro.py", line 5, in f
    time.sleep(0.5)  # to be sure this frame gets sampled
  File "/home/snoopjedi/repos/scalene/scalene/scalene_profiler.py", line 488, in cpu_signal_handler
    Scalene.cpu_signal_handler_helper(signum, this_frame)
  File "/home/snoopjedi/repos/scalene/scalene/scalene_profiler.py", line 604, in cpu_signal_handler_helper
    Scalene.enter_function_meta(frame, Scalene.__stats)
  File "/home/snoopjedi/repos/scalene/scalene/scalene_profiler.py", line 775, in enter_function_meta
    prepend_name = f.f_locals["cls"].__name__
AttributeError: 'str' object has no attribute '__name__'
Scalene: Program did not run for long enough to profile.
```

**Expected behavior**
Scalene proceeds as if the name was not defined, i.e. without prepending anything to `f_name`

 - OS: Ubuntu 20.04
 - Version `scalene 1.3.4.7`
 - I have observed the fault on a non-editable install of the library as well as the editable install shown above.

**Additional context**
Failure was first observed when profiling a program that indirectly invoked [`matplotlib.font_manager._json_decode()`](https://github.com/matplotlib/matplotlib/blob/69b51f1f4e7f8bab684fdf925ef8e2f94bbf2031/lib/matplotlib/font_manager.py#L932-L947)

It looks like cd5b7e7 may have been intended to address this problem. I can't find an associated issue.

SnoopJeDi:
In addition to using `getattr()` instead of dot notation, this PR also removes the exception handler introduced by cd5b7e7, because it seems like a `KeyError` could only originate here if the `f_locals` dict changes between the membership test and the indexing operation. That _might_ be possible, but would be better addressed by accessing `f_locals` only once and caching the result.

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -771,14 +771,11 @@ class Scalene:
                     fn_name = prepend_name + "." + fn_name
                 break
             if "cls" in f.f_locals:
-                try:
-                    prepend_name = f.f_locals["cls"].__name__
-                    if "Scalene" in prepend_name:
-                        break
-                    fn_name = prepend_name + "." + fn_name
+                prepend_name = getattr(f.f_locals["cls"], "__name__", None)
+                if not prepend_name or "Scalene" in prepend_name:
                     break
-                except KeyError:
-                    pass
+                fn_name = prepend_name + "." + fn_name
+                break
             f = f.f_back
 
         stats.function_map[fname][lineno] = fn_name
--- a/test/issues/test-issue193.py
+++ b/test/issues/test-issue193.py
@@ -1,0 +1,9 @@
+import time
+
+def test_cls_in_locals():
+    cls = "This value is not a class"
+    time.sleep(0.5)
+
+
+if __name__ == "__main__":
+    test_cls_in_locals()


emeryberger:
Thanks for the PR!

Issue #190, opened 2021-05-18
Scalene influences memory usage of python script


lenardkoomen-fin:
**Describe the bug**
I notice a much higher memory usage of my python script when running it with the `scalene` compared to running it with the normal `python` command. Also, the memory usage when using Scalene fluctuates wildly per run.

**To Reproduce**
A minimal python script:
```
import numpy as np

def main():
    n = 1000
    A, B = np.zeros((n, n)), np.zeros((n, n))
    for _ in range(1000):
        C = A @ B
    return C

if __name__ == "__main__":
    main()
```

When calling scalene like so: `$ scalene path/to/main.py`, the memory usage jumps to above 1 GB while the actual usage when calling the script normally with the `python` command is around 55 MB (as validated by the `memory-profiler` package and by looking at the macOS activity monitor). Also, the memory usage reported by Scalene fluctuates wildly per run.

**Expected behavior**
Scalene should provide a memory usage estimate equal to the actual usage of the Python script without influencing it. Also, the estimate should probably not fluctuate so much per run.

**Version info**
 - OS: macOS Catalina 10.15.6
- Python 3.9.1
- Scalene 1.3.4.7
- Numpy 1.20.3

jaltmayerpizzorno:
Hi, thank you for reaching out. How are you reading the memory usage in the macOS activity monitor -- simply the "Memory" column in the "Memory" tab (other columns are available, such as "Real Mem"...).

Scalene on MacOS does currently prevent Python's internal `pymalloc` allocator from being used, which is bound to have an impact on the memory footprint. We are hoping to change this in a future version.

emeryberger:
Hi @lenardkoomen-fin - thanks for the report! We tracked it down to the way Scalene was handling signals, which we were already looking at replacing. I have verified that this dramatically reduces the scope of the reported problem (which, FWIW, was due to thread stack creation). Memory still seems to be growing, but nowhere near as fast (still something to look into).

lenardkoomen-fin:
@jaltmayerpizzorno I interpreted the "net" column as being the net memory usage, and have also validated that figure using the macOS activity monitor.

@emeryberger No problem. I'll try to test whether this has solved the problem on my end ASAP.

emeryberger:
Confirmed fixed by others, closing. @lenardkoomen-fin if this is inaccurate, please re-open - thanks!

PR #202, opened 2021-05-27
Added initial queue handling (one thread per signal type).

emeryberger:


emeryberger:
Fixes https://github.com/plasma-umass/scalene/issues/190

PR #203, opened 2021-05-27
F202105 signal queue

jaltmayerpizzorno:
This is the PR I have been working on, but which has been mostly superseded by Emery's queueing code.


jaltmayerpizzorno:
Please review, but I think it's good to merge now.

scalene/scalene_sigqueue.py
```diff
@@ -14,15 +15,19 @@ def get(self):
...
+        if self.thread != None:
+            self.queue.put(None)
+            self.thread.join()
+            self.thread = None
 
     def run(self):
         while True:
             item = self.queue.get()
-            if not item:
+            if item == None:  # None == stop request
```
emeryberger, line 31: It's "Pythonic" to treat falsy things like `None` as bools.

scalene/scalene_sigqueue.py
```diff
@@ -14,15 +15,19 @@ def get(self):
         return self.queue.get()
 
     def start(self):
+        assert self.thread == None
```
emeryberger, line 18: `assert not self.thread`

scalene/scalene_profiler.py
```diff
@@ -1219,12 +1217,10 @@ def disable_signals() -> None:
...
             signal.signal(ScaleneSignals.free_signal, signal.SIG_IGN)
             signal.signal(ScaleneSignals.memcpy_signal, signal.SIG_IGN)
-            Scalene.__alloc_sigq.stop()
-            Scalene.__memcpy_sigq.stop()
-            Scalene.__cpu_sigq.stop()
+            Scalene.stop_signal_queues()
         except BaseException:
             # Retry just in case we get interrupted by one of our own signals.
-            Scalene.disable_signals()
+            Scalene.disable_signals()   # FIXME this could loop with an error
```
emeryberger, line 1223: If this could really loop with an error, shouldn't we handle it?

scalene/scalene_sigqueue.py
jaltmayerpizzorno: I thought checking for `None` explicitly was clearer, as we're looking for a specific value, rather than simply "anything that evaluates `False`"

scalene/scalene_sigqueue.py
```diff
@@ -14,15 +15,19 @@ def get(self):
...
         self.thread = threading.Thread(target=self.run, daemon=True)
         self.thread.start()
 
     def stop(self):
-        self.queue.put(None)
+        if self.thread != None:
+            self.queue.put(None)
+            self.thread.join()
```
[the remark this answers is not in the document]
jaltmayerpizzorno: To avoid possible lock corruption, need a guarantee that the thread is done before we can `fork`.

scalene/scalene_profiler.py
jaltmayerpizzorno: Sure, it wasn't my purpose with the branch, but I can do it.

scalene/scalene_sigqueue.py
jaltmayerpizzorno: Yup, thanks.

scalene/scalene_sigqueue.py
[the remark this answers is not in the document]
jaltmayerpizzorno: We could, in fact, create the threads with `daemon=False` (the default). It's a bit of defensive programming to declare them daemons (if we miss stopping them, nothing hangs).

scalene/scalene_sigqueue.py
emeryberger: Any actual value will be truthy, so I don't think this is a problem. In principle, we should use a type annotation for this. But it's fine.

scalene/scalene_sigqueue.py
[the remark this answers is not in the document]
emeryberger: Yes, let's keep them as daemons to avoid hangs. Please add a comment explaining the reason for the join.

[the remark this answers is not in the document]
jaltmayerpizzorno: Done!

emeryberger (approved):

PR #210, opened 2021-06-02
Fixes libscalene preloading, DRYes some code duplications

jaltmayerpizzorno:
Note that `libscalene.dylib` and `libscalene.so` are no longer generated at the top level directory, but directly into `scalene`. The homebrew formula may need to be adjusted.

### DIFF
--- a/scalene/scalene_preload.py
+++ b/scalene/scalene_preload.py
@@ -51,9 +51,6 @@ class ScalenePreload:
                 "Scalene warning: currently only 64-bit x86-64 and ARM platforms are supported for memory and copy profiling."
             )
 
-        # Load shared objects (that is, interpose on malloc, memcpy and friends)
-        # unless the user specifies "--cpu-only" at the command-line.
-
         try:
             from IPython import get_ipython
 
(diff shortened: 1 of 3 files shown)


scalene/scalene_preload.py
```diff
@@ -31,9 +54,6 @@ def setup_preload(args: argparse.Namespace) -> bool:
         # Load shared objects (that is, interpose on malloc, memcpy and friends)
         # unless the user specifies "--cpu-only" at the command-line.
 
-        if args.cpu_only:
```
emeryberger, line 34: Right now, we need this for Jupyter notebooks since preloading doesn't yet work there.

scalene/scalene_preload.py
```diff
@@ -43,15 +63,12 @@ def setup_preload(args: argparse.Namespace) -> bool:
...
-        if sys.platform == "linux":
-            if ("LD_PRELOAD" not in os.environ) and (
-                "PYTHONMALLOC" not in os.environ
-            ):
-                os.environ["LD_PRELOAD"] = os.path.join(
-                    os.path.dirname(scalene.__path__[0]), "libscalene.so"
-                )
-                os.environ["PYTHONMALLOC"] = "malloc"
+        # Start a subprocess with the required environment variables.
+        if sys.platform == "linux" or sys.platform == "darwin":
```
emeryberger, line 67: AFAICT this test is redundant (already in the `get_preload_environ` call).

emeryberger (changes requested):
Overall looks good, just a few comments / fixes.

scalene/scalene_preload.py
jaltmayerpizzorno: That check is still there, in `get_preload_environ()`.  We don't set up the preloading environment variables if `cpu_only`, and if no environment variables need overriding, we don't start a new process.

On MacOS, we need python to be started with `OBJC_DISABLE_INITIALIZE_FORK_SAFETY` defined to support multiprocessing.

jaltmayerpizzorno: BTW, I tested with Jupyter on my mac, and the memory profiling seemed to work...

scalene/scalene_preload.py
jaltmayerpizzorno: Right, I'll remove. I thought of it, but was thinking we'll need to add something different for Windows support.

scalene/scalene_preload.py
emeryberger: Wow, that'd be awesome.

emeryberger (approved):

PR #228, opened 2021-06-17
F202106 GitHub release workflow

jaltmayerpizzorno:
This fixes some issues with setup.py, adds "make pkg" for packaging scalene, adds x86_64 -> arm64 cross building, and adds a github action to build and upload scalene to pypi upon releases.

The arm64 and github action still needs testing, but I'm opening the PR so that it can be reviewed when convenient.


jaltmayerpizzorno:
I think this PR is ready to go now.

Issue #250, opened 2021-08-09
TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'


rmcgibbo:
**Describe the bug**
I got a traceback `TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'` while running profiling.

**To Reproduce**
Unfortunately, the code is proprietary so I can't share a full repo.

**Screenshots**
![image](https://user-images.githubusercontent.com/641278/128752560-9fb9975f-4ea4-4c18-a5c9-20ca5f7ed3ca.png)

**Desktop (please complete the following information):**
 - OS: CentOS 7 Linux
- Python: 3.7
- Scalene 2.5.2

**Additional context**
It's possible that the issue is that while the code was running under the profiler, I may have been (okay, I was) poking it with gdb and looking at the backtrace, and then quitting gdb.

emeryberger:
Hi - 2.5.2 is not an actual version number for Scalene, which is 1.3.12 (at this writing). Can you clarify? Also, IIRC Scalene requires Python 3.8 for Windows. Please try installing via `pip` and let us know what happens.

rmcgibbo:
Sorry, it looks like it's actually `1.3.12`. We appeared to just made an error in naming it in our system. But yes, it is python3.7 on Linux.

emeryberger:
We now have an `assert` in place that should be triggered if this situation ever arises again (it never should...). You can update with `python3 -m pip install git+https://github.com/plasma-umass/scalene`. I suspect that gdb might be the culprit in that it could be interfering with the signals that Scalene relies on (these are all here: https://github.com/plasma-umass/scalene/blob/master/scalene/scalene_signals.py).

rmcgibbo:
Sounds good. Thanks!

PR #251, opened 2021-08-10
Changes to get fewer errors from `mypy`

emeryberger:
Still not mypy-clean, but it turned up the mismatch that gets reported by https://github.com/plasma-umass/scalene/issues/250, where for some reason it doesn't look like `ScaleneStatistics.start_clock` is ever invoked. Moved some stuff to use `pathlib` which is cleaner than using strings and avoids the need to cast the alias directory to `Filename` (which was never quite right anyway).

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -66,11 +66,14 @@ from scalene.scalene_gpu import ScaleneGPU
 from scalene.scalene_parseargs import ScaleneParseArgs, StopJupyterExecution
 from scalene.scalene_sigqueue import ScaleneSigQueue
 
-def require_python(version: Tuple[int, int]) -> None:
-    assert (sys.version_info >= version), \
-           f"Scalene requires Python version {version[0]}.{version[1]} or above."
 
-require_python((3,7) if sys.platform != "win32" else (3,8))
+def require_python(version: Tuple[int, int]) -> None:
+    assert (
+        sys.version_info >= version
+    ), f"Scalene requires Python version {version[0]}.{version[1]} or above."
+
+
+require_python((3, 7) if sys.platform != "win32" else (3, 8))
 
 
 # Scalene fully supports Unix-like operating systems; in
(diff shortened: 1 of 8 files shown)


scalene/scalene_profiler.py
```diff
@@ -224,9 +235,16 @@ def get_original_lock() -> threading.Lock:
...
+    __is_thread_sleeping: Dict[int, bool] = defaultdict(
+        bool
+    )  # False by default
     __child_pids: Set[int] = set()
 
+    # Signal queues for CPU timers, allocations, and memcpy
+    __cpu_sigq: ScaleneSigQueue[Any]
+    __alloc_sigq: ScaleneSigQueue[Any]
+    __memcpy_sigq: ScaleneSigQueue[Any]
+
```
jaltmayerpizzorno, line 247: Maybe set these to `None` here?

scalene/scalene_statistics.py
```diff
@@ -32,7 +32,7 @@ class ScaleneStatistics:
     #
     def __init__(self) -> None:
         # time the profiling started
-        self.start_time: float = None
+        self.start_time: float = 0
```
jaltmayerpizzorno, line 35: Leaving it at `None` would help detect cases where `start_clock` wasn't called.

scalene/scalene_statistics.py
```diff
@@ -194,7 +194,7 @@ def start_clock(self) -> None:
 
     def stop_clock(self) -> None:
         self.elapsed_time += time.perf_counter() - self.start_time
-        self.start_time = None
+        self.start_time = 0
```
jaltmayerpizzorno, line 197: Same as above, about `None`.

jaltmayerpizzorno (commented):
I suggest we commit reformatter passes separately, to reduce noise from when actual changes are done.

scalene/scalene_profiler.py
emeryberger: I didn't do that because then they have to be `Optional` types, and anyway they get initialized by `__init__`.

scalene/scalene_statistics.py
emeryberger: This is true, but it currently fails with an exception, _and_ it would force us to use an `Optional` type. I'd rather handle it without breaking the typesystem.

scalene/scalene_statistics.py
emeryberger: Same as above, about `Optional` 😄

scalene/scalene_statistics.py
jaltmayerpizzorno: How about an assertion that `start_time != 0`?  Otherwise we'll silently use a large elapsed time should `start_time` not be invoked.

emeryberger: It honestly should never happen, so I guess an assert is the right choice. (I don't know why or how it happened for the mentioned issue.)

PR #259, opened 2021-08-23
Line level memory attribution

emeryberger:
Memory allocation attribution is now to the specific file and line number of Python, rather than sampled. This avoids the problem of "slippage" of attribution of allocations, giving much crisper results.


emeryberger:
For clarity: we are still using sampling of allocations, but the sampling has been corrected (in terms of number of bytes attributed, and the specific location rather than whatever is running when the signal is received).

emeryberger:
Not currently working on Linux because of an initialization / malloc interposition failure.

PR #271, opened 2021-09-25
Changed iterations to generators, removed redundant check.

emeryberger:
Replaced some loops with generators (cleaner and perhaps faster). Removed a redundant check which became obvious once this was done (in `should_trace` - we were checking the condition computed for `found_in_profile_only` twice).


scalene/scalene_profiler.py
```diff
@@ -1189,20 +1187,11 @@ def should_trace(filename: str) -> bool:
...
-
+        found_in_profile_only = any(prof in filename for prof in profile_only_list)
         if not found_in_profile_only:
             return False
         if Scalene.__args.profile_all:
             # Profile everything else, except for "only" choices.
-            found_in_profile_only = False
-            for prof in profile_only_list:
-                if prof in filename:
-                    return True
```
sternj, line 1205: Concerned about this logic here-- wasn't "profile-only" at one point used to exclude files, but that logic doesn't seem to be used anymore. What about that short circuit that used to be here?

sternj (changes requested):

scalene/scalene_profiler.py
emeryberger: It is still there - it's just tested once now, instead of twice.

emeryberger: By the time we get to this line, `found_in_profile_only` must be true.

sternj: ok so I see that a particular optimization is unneeded because that loop has also been exorcised, but this still doesn't "profile everything else, except for "only" choices" (though it did not before)-- has this behavior been deprecated?

emeryberger: Yeah, that is I think just a buggy comment!

emeryberger: It should probably read "profile everything as long as it matches the profile-only strings, if any."

emeryberger: It should probably read "profile everything as long as it matches the profile-only strings, if any."

scalene/scalene_profiler.py
```diff
@@ -1189,20 +1187,11 @@ def should_trace(filename: str) -> bool:
...
             if profile_only_list and not found_in_profile_only:
```
jaltmayerpizzorno, line 1193: As it is now, `not found_in_profile_only` must be `True` at this point.

scalene/scalene_profiler.py
```diff
@@ -1189,20 +1187,11 @@ def should_trace(filename: str) -> bool:
...
             return False
-        found_in_profile_only = False
-        for prof in profile_only_list:
-            if prof in filename:
-                found_in_profile_only = True
-                break
-
+        found_in_profile_only = any(prof in filename for prof in profile_only_list)
         if not found_in_profile_only:
             return False
```
jaltmayerpizzorno, line 1190: Shouldn't this be
```python
        if profile_only_list and not any(prof in filename for prof in profile_only_list):
            return False

        if Scalene.__args.profile_all:
            # Profile everything else, except for "only" choices.
            return True
```

emeryberger: I think the logic should be as follows:
* If `profile-only` is specified, exclude files that don't match any provided pattern.

I am committing a change that I hope simplifies this and makes it clear.

scalene/scalene_profiler.py
emeryberger: Good point, fixed.

PR #281, opened 2021-10-11
Added missing includes and removed unused headers from src/include

superbobry:
Hi there,

This is a controversial PR and I'm okay with it not being merged as-is. However, I thought I'd start a discussion about making Scalene a bit easier to navigate for contributors. Having dead code around is a bit, and unless it is only temporarily dead, I think it'd be best to remove it altogether. Wdyt?

On a slightly unrelated note, thank you for building scalene! I've watched the PyCon talk, and I'm really excited that the Python ecosystem now has a low overhead all-in-one tool for diagnosing performance issues.


emeryberger:
LGTM - thanks!

PR #295, opened 2021-11-15
Added function summary field to JSON

sternj:
created 'functions' block with all function summary information


emeryberger:
Merging. Please update `scalene_output.py` to remove redundant work.

PR #306, opened 2021-12-22
Fixes incompatibility with rich < 9.4.0

nixjdm:
Without this you'd get a

```
TypeError: add_row() got an unexpected keyword argument 'end_section'
```
The output with this diff on rich 9.2.0 still looks fine to me.

[Airflow 2.0.0 for instance still requires rich 9.2.0](https://github.com/apache/airflow/blob/2.0.0/setup.cfg#L129), for instance, so I'd say there's still reason to support rich down to that version.

### DIFF
--- a/scalene/scalene_output.py
+++ b/scalene/scalene_output.py
@@ -530,7 +530,10 @@ class ScaleneOutput:
                 break
 
             if print_fn_summary:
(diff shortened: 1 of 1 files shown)


emeryberger:
Thanks! (PR merged outside Halcyon in Austin :).

PR #316, opened 2022-01-08
Changed logic to get sizes from headers

sternj:
- Changed call to getSize to treat all objects as headered
- Changed reporting to report on requested size, rather than given size (Potential issue, but strengthens against apparent volatility of getSize)


sternj:
This did not work, inconsistencies were due to initialization that was being picked up.

PR #317, opened 2022-01-08
Factor out Header class

sternj:


emeryberger (changes requested):
Go ahead and remove the commented-out stuff from `libscalene.cpp`.

sternj:
This was fixing an issue that did not exist

sternj:
Reopening to rework for refactor

emeryberger (approved):
Minor comment - it's good practice to switch to using the more modern `#pragma once` in header files, which solves the problem that the `#ifndef` two-step handles.

PR #332, opened 2022-01-20
Update bm_scimark.py

zjzh:
refactoring code with var unpack which is more pythonic, concise, readable and efficient; how do think this change which has practical value？


emeryberger:
Thanks - this benchmark comes from the [Python pyperformance suite](https://github.com/python/pyperformance/tree/main/pyperformance/data-files/benchmarks), so we don't want to make changes to it.

Issue #201, opened 2021-05-26
Nested lists comprehension throws error


joaopmatias:
Hi!
I started using scalene recently and saw an unexpected behavior.

Thank you in advance!

**Describe the bug**
The following Python command throws an error.
```
n = 100_000
nested = [[0 for _ in range(n)] for _ in range(1)]
print("Finished", flush=True)
```

**To Reproduce**
1. Write the following code to a file `nested.py`
```
n = 100_000
nested = [[0 for _ in range(n)] for _ in range(1)]
print("Finished", flush=True)
```
2. Run `scalene nested.py` in the terminal.
3. If an error is not thrown, run the script multiple times in a row or change `n`, for example, `n=1_000_000`.

**Expected behavior**
Repeated error messages are printed. Portions of those messages include
```
[...]
Error in program being profiled:
 maximum recursion depth exceeded
Traceback (most recent call last):
  File "/opt/miniconda3/envs/tresnove/lib/python3.9/site-packages/scalene/scalene_profiler.py", line 1237, in profile_code
    exec(code, the_globals, the_locals)
  File "sca.py", line 2, in <module>
    nested = [[0 for _ in range(n)] for _ in range(1)]
  File "sca.py", line 2, in <listcomp>
    nested = [[0 for _ in range(n)] for _ in range(1)]
  File "sca.py", line 2, in <listcomp>
    nested = [[0 for _ in range(n)] for _ in range(1)]
  File "/opt/miniconda3/envs/tresnove/lib/python3.9/site-packages/scalene/scalene_profiler.py", line 796, in malloc_signal_handler
    Scalene.allocation_signal_handler_helper(signum, this_frame, "malloc")
  File "/opt/miniconda3/envs/tresnove/lib/python3.9/site-packages/scalene/scalene_profiler.py", line 822, in allocation_signal_handler_helper
    t.start()
  File "/opt/miniconda3/envs/tresnove/lib/python3.9/threading.py", line 879, in start
    self._started.wait()
  File "/opt/miniconda3/envs/tresnove/lib/python3.9/threading.py", line 574, in wait
    signaled = self._cond.wait(timeout)
  File "/opt/miniconda3/envs/tresnove/lib/python3.9/threading.py", line 312, in wait
    waiter.acquire()
  File "/opt/miniconda3/envs/tresnove/lib/python3.9/site-packages/scalene/scalene_profiler.py", line 490, in cpu_signal_handler
[...]

PR #335, opened 2022-01-24
Fixed scalene looping infinitely in some functions

sternj:
Fixed an anomalous loop condition that I suspect caused #201 as well as several other bits of weirdness. Logic was supposed to be "If code in a special block such as `<module>` or `<listcomp>`, keep going back in the stack until you find real code", but this logic has been around for a long time.

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -937,8 +937,10 @@ class Scalene:
         f = frame
         try:
             while "<" in Filename(f.f_code.co_name):
-                f = cast(FrameType, f.f_back)
-                
+                f = cast(FrameType, frame.f_back)
+                # Handle case where the <\w+> is at the bottom of the stack
+                if f is None:
+                    return
         except:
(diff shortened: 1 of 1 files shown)


scalene/scalene_profiler.py
```diff
@@ -937,12 +937,9 @@ def enter_function_meta(
...
         try:
             while "<" in Filename(f.f_code.co_name):
-                f = cast(FrameType, frame.f_back)
-                if (
-                    "<genexpr>" in f.f_code.co_name
-                    or "<module>" in f.f_code.co_name
-                    or "<listcomp>" in f.f_code.co_name
-                ):
+                f = cast(FrameType, f.f_back)
+                # Handle case where the <\w+> is at the bottom of the stack
```
emeryberger, line 941: This comment + regexp is pretty opaque. How about "where the function named <whatever> ...".

emeryberger (approved):

PR #341, opened 2022-02-02
Added --port cli argument

legau:
This PR adds a --port argument to specify the port bound by the local web server


emeryberger:
Thanks for the PR! Merging. Related question: is there a better default port number?

PR #347, opened 2022-02-08
Added support for profiling programs that use signals

sternj:
The current Scalene implementation breaks when anyone attempts to use a signal that we use. This blocks any client program from using a signal that we use.

However, the exception here is the alarm signal (`SIGALRM` or `SIGVTALRM`), which is routed to another signal (`SIGUSR1` now, but this likely has to change). The current approach hijacks the timer request from the user program and delivers the user timer using this alternate signal while maintaining its current timer loop in the background. 

This is still a WIP and needs extensive testing because of how many corner cases there are.


scalene/scalene_client_timer.py
```diff
@@ -0,0 +1,46 @@
+from typing import Tuple
+
+
+class ScaleneClientTimer:
+    seconds: float
```
jaltmayerpizzorno, line 5: This could use a short description.

scalene/scalene_client_timer.py
```diff
@@ -0,0 +1,46 @@
...
+        self.is_set = True
+
+    def reset(self):
+        self.seconds = 0.0
+        self.interval = 0.0
+        self.is_set = False
+    def get_itimer(self) -> Tuple[float, float]:
+        return self.seconds, self.interval
+
+    def yield_next_delay(self, elapsed) -> Tuple[bool, float]:
```
jaltmayerpizzorno, line 33: This could use a description

scalene/scalene_client_timer.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class ScaleneClientTimer:
+    seconds: float
+    interval: float
+    remaining_seconds: float
+    remaining_interval: float
+    delay_elapsed: bool
+
+    is_set: bool
```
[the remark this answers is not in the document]
sternj: [yes!](https://docs.python.org/3/library/typing.html#typing.ClassVar)

PR #371, opened 2022-03-06
Add override of realloc to prevent reentrancy 

### LINKED ISSUE #360: Scalene SIGSEGV
Possibly a replication of #354 . A SIGSEGV arises during the "copy" phase of a realloc. Still working on tracking down the origin of the issue, but the SIGSEGV occurs at `memcpysampler.hpp:45`. To run this in Docker, allocate 3GiB of memory to the docker engine-- it normally segfaults by iteration 19.

[Code for replication located here](https://gist.github.com/sternj/a031f8c084e7312146902ae21c2efda6)

sternj:
Fixes #360 -- A highly network-bound program was running into `SIGSEGV`s. Inspection in `gdb` revealed that this was a double free happening in the arena allocator. In the process of adding a new arena, the array of arenas was being reallocated with `PyMem_RawRealloc`, and because `realloc` was previously overloaded as a composition of `malloc` and `free`. At some point in the middle of this, `SampleHeap::process_malloc` was being called, which called `whereInPython`, which queries the Python interpreter, which in turn allocates memory and creates a new arena, which requires expanding the list of arenas. This means that the list being `realloc`ed is freed in the middle of the call to `realloc`, leading to a double free. The solution to this was to [add the ability to redefine `realloc` in Heap-Layers](https://github.com/emeryberger/Heap-Layers/pull/45) and then redefine `realloc` in `SampleHeap`. We verified that this resolves the issue by running the test program in a Vagrant instance running Ubuntu 21.10.

### DIFF
--- a/src/include/sampleheap.hpp
+++ b/src/include/sampleheap.hpp
@@ -40,14 +40,6 @@ static SampleFile& getSampleFile() {
 }
 
 #define USE_ATOMICS 0
-#define PRINT_ALL 0
-#define MEMOIZE_LAST 1
-#if MEMOIZE_LAST
-void* last_call[20];
-int n_bt = 0;
-// char** last_call_s;
-void* last_freed;
-#endif 
 #if USE_ATOMICS
 typedef std::atomic<uint64_t> counterType;
 #else
@@ -90,9 +82,6 @@ class SampleHeap : public SuperHeap {
   ATTRIBUTE_ALWAYS_INLINE inline void* malloc(size_t sz) {
     MallocRecursionGuard g;
     auto ptr = SuperHeap::malloc(sz);
-    #if PRINT_ALL
-    printf_("=== MI %p\n", ptr);
-    #endif
     if (unlikely(ptr == nullptr)) {
       return nullptr;
     }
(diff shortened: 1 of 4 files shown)


src/include/sampleheap.hpp
```diff
@@ -149,19 +201,25 @@ class SampleHeap : public SuperHeap {
...
     }
-    SuperHeap::free(ptr);
+    
+    
+    
+    
   }
 
   inline void register_free(size_t realSize, void* ptr) {
+    // printf_("In realloc %p\n", &in_realloc);
```
emeryberger, line 220: Leftover comment cruft.

src/include/scaleneheader.hpp
```diff
@@ -28,12 +28,13 @@ class ScaleneHeader {
 #if USE_HEADERS
 #if DEBUG_HEADER
   ScaleneHeader(size_t sz) : size(sz), magic(MAGIC_NUMBER) {}
-  alignas(std::max_align_t) size_t size;
+  size_t size;
+  size_t padding;
   size_t magic;
 #else
   ScaleneHeader(size_t sz) : size(sz) {}
+  // size_t size;
```
emeryberger, line 36: Leftover comment cruft.

src/include/scaleneheader.hpp
```diff
@@ -52,6 +53,9 @@ class ScaleneHeader {
   static inline size_t getSize(void *ptr) {
 #if USE_HEADERS
 #if DEBUG_HEADER
+    if(getHeader(ptr)->magic != MAGIC_NUMBER) {
```
emeryberger, line 56: Leftover warning cruft.

emeryberger (changes requested):
Remove some leftover cruft

PR #377, opened 2022-03-28
Fix issue where memory is not being reported on Linux

sternj:
Reworks memory accounting so that memory whose allocation initiates in Python is attributed to Python. Also (combined with a fix to heaplayers) fixes an issue where memory originating in Python is counted differently on Mac and Linux or is simply not counted on Linux at all. 

Still a WIP-- counting for lines is slightly different and I want to figure out where that's coming from


src/include/sampleheap.hpp
```diff
@@ -86,6 +86,7 @@ class SampleHeap : public SuperHeap {
       return nullptr;
     }
     if (pythonDetected() && !g.wasInMalloc()) {
+      // printf_("getting size from %p ... ", ptr);
```
emeryberger, line 89: FYI lots of comments left in

src/include/heapredirect.h
```diff
@@ -0,0 +1,274 @@
+/*
```
emeryberger, line 1: Whole file should probably not be replaced?

sternj: I'm quite confused as to how this happened, but `diff` says it's the same as the one in `vendor/Heap-Layers/wrappers` so I'm gonna feel safe deleting it

src/include/sampleheap.hpp
```diff
@@ -136,13 +135,10 @@ class SampleHeap : public SuperHeap {
     SuperHeap::free(ptr);
     if (buf) {
       if (sz < buf_size) {
-        if (buf_size - sz <= PYMALLOC_MAX_SIZE) {
```
sternj, line 139: This was an oversight that I made in previous commits, this line does not belong here

src/include/sampleheap.hpp
```diff
@@ -252,7 +246,7 @@ class SampleHeap : public SuperHeap {
       auto realSize = SuperHeap::getSize(ptr);
       assert(realSize >= sz);
       assert((sz < 16) || (realSize <= 2 * sz));
-      register_malloc(realSize, ptr);
```
sternj, line 255: Another oversight, this optional boolean indicates that it is not from pymalloc

src/source/libscalene.cpp
```diff
@@ -141,6 +141,7 @@ class MakeLocalAllocator {
   }
 
   static inline void *local_malloc(void *ctx, size_t len) {
+    MallocRecursionGuard m;
```
sternj, line 144: Recursion guard added here because we only want memory accounting at the top level-- if it's passed from `local_malloc` into the system malloc, we still want to attribute it here.

src/source/libscalene.cpp
```diff
@@ -158,22 +159,19 @@ class MakeLocalAllocator {
     const auto allocSize = len + sizeof(ScaleneHeader);
     buf = get_original_allocator()->malloc(ctx, allocSize);
     auto *header = new (buf) ScaleneHeader(len);
+    class Nada {};
 #else
     auto *header = (ScaleneHeader *)get_original_allocator()->malloc(ctx, len);
 #endif
     assert(header);                  // We expect this to always succeed.
-    if (len <= PYMALLOC_MAX_SIZE) {  // don't count allocations pymalloc passes
+    if (! m.wasInMalloc() || len == SampleHeap<1, HL::NullHeap<Nada>>::NEWLINE ) {  // don't count allocations pymalloc passes
```
sternj, line 167: An attempt at a check for if we were missing `NEWLINE`s, should be removed

emeryberger: @sternj still here?

PR #381, opened 2022-03-31
make normal SIGTERM exit cleanly

sternj:
Workers of `multiprocessing.Pool` were not writing out profiling information in Linux. Further investigation revealed that it is because MacOS and Linux handle SIGTERM differently, so we made the behavior consistent by adding a signal handler that cleanly exits without yielding control flow back to the worker loop

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -395,6 +395,16 @@ class Scalene:
             sigq.stop()
 
     @staticmethod
+    def term_signal_handler(
+        signum: Union[
+            Callable[[Signals, FrameType], None], int, Handlers, None
+        ],
+        this_frame: Optional[FrameType],
+    ):
+        raise SystemExit(143)
+
+
+    @staticmethod
     def malloc_signal_handler(
         signum: Union[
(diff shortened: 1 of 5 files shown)


scalene/scalene_profiler.py
```diff
@@ -394,6 +394,20 @@ def stop_signal_queues() -> None:
...
+        signum: Union[
+            Callable[[Signals, FrameType], None], int, Handlers, None
+        ],
+        this_frame: Optional[FrameType],
+    ):
+        
+        Scalene.stop()
+        Scalene.output_profile()
+        
+        Scalene.__orig_exit(143)
```
emeryberger, line 408: https://en.wikipedia.org/wiki/Magic_number_(programming)

scalene/scalene_profiler.py
```diff
@@ -1540,6 +1558,7 @@ def profile_code(
             # Intercept sys.exit and propagate the error code.
             exit_status = se.code
         except KeyboardInterrupt:
+            traceback.print_exc()
```
emeryberger, line 1561: Leftover debugging artifact?

emeryberger (changes requested):

PR #393, opened 2022-05-05
Change function to prevent memory smearing in 3.10

sternj:


sternj:
Note that time smearing is still a significant issue, but one outside of the scope of this PR

PR #399, opened 2022-06-01
Initial 3 11 support

emeryberger:
Changed to use new 3.11 frame ABI (see https://docs.python.org/3.11/whatsnew/3.11.html#pyframeobject-3-11-hiding).

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -268,14 +268,14 @@ class Scalene:
         of code or not in invalidate_lines (for per-line memory
         accounting).
         """
-        found_frame = None
         f = frame
+        current_file_and_line = (fname, lineno)
         while f:
-            if (f.f_code.co_filename, f.f_lineno) == (fname, lineno):
-                found_frame = f
+            if (f.f_code.co_filename, f.f_lineno) == current_file_and_line:
+                return f
                 break
             f = cast(FrameType, f.f_back)
-        return found_frame
+        return None
 
     @staticmethod
     def update_line() -> None:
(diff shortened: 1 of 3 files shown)


scalene/scalene_profiler.py
```diff
@@ -268,14 +268,14 @@ def on_stack(
...
         """
-        found_frame = None
         f = frame
+        current_file_and_line = (fname, lineno)
         while f:
-            if (f.f_code.co_filename, f.f_lineno) == (fname, lineno):
-                found_frame = f
+            if (f.f_code.co_filename, f.f_lineno) == current_file_and_line:
+                return f
                 break
```
jaltmayerpizzorno, line 276: Do we still need this break?

setup.py
```diff
@@ -159,7 +159,7 @@ def build_libscalene(self):
     install_requires=[
         "rich>=9.2.0",
         "cloudpickle>=1.5.0",
-        "numpy",
+#        "numpy",
```
jaltmayerpizzorno, line 162: Maybe plain delete this?

src/source/pywhere.cpp
```diff
@@ -159,6 +159,23 @@ static PyThreadState* findMainPythonThread() {
   return main;
 }
 
+#if !((PY_MAJOR_VERSION == 3) && (PY_MINOR_VERSION >= 11))
+  // If we aren't compiling for version 3.11 or higher, define
+  // replacements for the new ABI.
+  inline PyFrameObject * PyThreadState_GetFrame(PyThreadState * threadState) {
+    return threadState->frame;
```
jaltmayerpizzorno, line 166: I don't think we need this wrapper -- it's been available a long time now.

jaltmayerpizzorno (changes requested):

setup.py
emeryberger: Agreed; this was tentatively commented out.

PR #406, opened 2022-07-01
Account for case wherein all threads are sleeping

sternj:


emeryberger (changes requested):
If `total_frames` is 0 but `new_frames` is not None, there will be an undefined variable error for all the averages, right?

PR #419, opened 2022-08-08
Hopefully fixed smart pointer class.

jaltmayerpizzorno:


sternj (approved):

src/include/pyptr.h
```diff
@@ -15,23 +15,25 @@ class PyPtr {
  public:
   PyPtr(O* o) : _obj(o) {}
 
-  O* operator->() { return _obj; }
+  PyPtr(const PyPtr& ptr) : _obj(ptr._obj) {
+    Py_IncRef((PyObject*)_obj);
+  }
 
-  operator O*() { return _obj; }
+   O* operator->() { return _obj; }
```
emeryberger, line 22: Please add an explanatory comment here (I understand why you did this but it should be documented).

src/include/pyptr.h
```diff
@@ -15,23 +15,25 @@ class PyPtr {
...
-  PyPtr& operator=(O* o) {
-    Py_DecRef((PyObject*)_obj);
-    _obj = o;
-    return *this;
-  }
+ explicit operator O*() { return _obj; }
 
-  PyPtr& operator=(PyPtr& ptr) {
+  PyPtr& operator=(const PyPtr& ptr) {
+    // increment first in case self-assigning
```
emeryberger, line 27: Because it could get GC'd if not, right? (Please elaborate either way)

emeryberger: Also, would an explicit equality check make sense instead?

emeryberger (changes requested):

src/include/pyptr.h
jaltmayerpizzorno: Yes -- if assigning to oneself, if you first decrement, you may lose the object. I'll add the equality check.

sternj: I mean an explicit check for equality with `nullptr` could make sense, but I think that the information that the `static_cast` provides to us as developers is helpful and is a good canary for if other changes will cause type issues

src/include/pyptr.h
sternj: I removed that operator actually, I just replaced it with a static cast

jaltmayerpizzorno: Sam got rid of that, so I think it's resolved.

emeryberger (approved):
LGTM.

Issue #425, opened 2022-08-26
Error getting real path: 2; Scalene error: received signal SIGABRT


ezyang:
**Describe the bug**
A program using importlib to import a module that imports request from urllib fails, when run with scalene.

**To Reproduce**
1. Create two folders: `benchmark/torchbenchmark` and `foo`
2. In `benchmark/torchbenchmark/__init__.py` put

```
from urllib import request
```

3. In `foo/c.py` put

```
import importlib
import os

os.chdir("../benchmark")
sys.path.append(os.getcwd())
module = importlib.import_module(f"torchbenchmark")
```

**Expected behavior**
Program passes with scalene

Actual behavor:
```
(/scratch/ezyang/pytorch-scratch2-env) ezyang@devfair040:/scratch/ezyang/torchdynamo$ python c.py
(/scratch/ezyang/pytorch-scratch2-env) ezyang@devfair040:/scratch/ezyang/torchdynamo$ scalene c.py
Error getting real path: 2
Scalene error: received signal SIGABRT
NOTE: The GPU is currently running in a mode that can reduce Scalene's accuracy when reporting GPU utilization.
Run once as root (i.e., prefixed with `sudo`) to enable per-process GPU accounting.
```

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: na
 - Version: na

emeryberger:
This should be fixed in the repository version: `python3 -m pip install git+https://github.com/plasma-umass/scalene`. (Please re-open if not!)

ezyang:
Looks like it is working now, thanks!

PR #426, opened 2022-08-27
Resolve paths according to original working directory.

emeryberger:
Right now, changing the current working directory in the middle of profiling disrupts name resolution (and even triggers an `abort`). This PR resolves all file names using the original invocation path.


emeryberger:
Fixes https://github.com/plasma-umass/scalene/issues/425.

Issue #397, opened 2022-05-19
Using scalene programatically fails with argument error


CatalinMustata:
**Describe the bug**
Trying to run scalene programmatically results in an error - this happens only when I use `from scalene import profile, scalene_profiler`.

```
usage: scalene.profile [-h] [--pid PID] (--on | --off)
scalene.profile: error: one of the arguments --on --off is required
Scalene: Program did not run for long enough to profile.
```

**To Reproduce**
- add `from scalene import profile, scalene_profiler`
- try to run program (`scalene --on test.py` or `python -m scalene --on test.py` yield the same result)

**Expected behavior**
For it to run, like it does without the import in my code.

**Desktop (please complete the following information):**
 - MacOS Monterey / Python 3.9.1
 - Scalene 1.5.8

emeryberger:
I can reproduce this but it works if you do not import `profile` (which you should not import!).

This example (`programmatic-test.py`) produces the expected results.

```Python
from scalene import scalene_profiler

x = 0
for i in range(10000):
    for j in range(1000):
        x += 1

# Turn profiling on
scalene_profiler.start()

x = 0
for i in range(10000):
    for j in range(1000):
        x += 1
        
# Turn profiling off
scalene_profiler.stop()
```

```shell
% python3 -m scalene --cli  --off ./programmatic-test.py
```

NivekT:
> I can reproduce this but it works if you do not import `profile` (which you should not import!).

Hi, I just found this issue because I ran into the same problem. I think some IDEs are prompting users to import `profile` with a red underline when `@profile` is used, so as a new user, I did. 

Perhaps nothing needs to be done about this, but if it is easy to tell users (or the IDEs) not to somewhere, it might be helpful.

Issue #244, opened 2021-07-21
Profiling python module


teristam:
**Is your feature request related to a problem? Please describe.**
I would like to profile my code as a module. I typically run it in command line like `python -m mypackage.mymodule`

**Describe the solution you'd like**
some addition flag to enable running as a module e.g. `scalene -m mypackage.mymodule`

**Describe alternatives you've considered**
We can create a wrapper script for the module and run it instead, but that's very inconvenient.

Thanks!

dlopes7:
If anyone arrives here like me, this is how you work around it:

```
import runpy

runpy.run_module('<module_name>', run_name='__main__')
```

PR #452, opened 2022-10-21
Add explanations about running scalene in README

laurentS:
This adds a few bits of explanations about running scalene, specifically for my use case :)
I struggled to get it running, and had to find #244 and #397 to figure out how to do it.

It's the 4th time I've been trying to use `scalene`, so I felt like a bit of docs couldn't hurt!


emeryberger:
Thanks!

PR #453, opened 2022-10-22
Use co_qualname instead of looping to find class info.

emeryberger:
This PR eliminates the current logic for producing fully-qualified function names. It does not behave _exactly_ the same way, in that it doesn't skip Scalene functions, but that seems superfluous.


sternj:
Perhaps might be good to tighten the `except` block to only handles `AttributeError` or whatever it would raise, other than that no objections!

PR #465, opened 2022-10-29
Fixed file handling

emeryberger:
Scalene now fails gracefully when it encounters a file name that does not correspond to a real file, like `<bootstrap something>`. Working on improving the handling the name of the input file to cope with invocations of files from a different directory than the current, or names containing `./` (for example).


emeryberger:
Seems to be working fine now.

PR #478, opened 2022-11-07
Updates for python 3, pylint, isort

marksmayo:
A bit of tidying up.  Updating code for python 3 (no need to inherit from object), as well as cleaning whitespace, sorting imports and other small improvements.

### DIFF
--- a/benchmarks/benchmark.py
+++ b/benchmarks/benchmark.py
@@ -1,9 +1,9 @@
 import os
-import sys
 import re
-import subprocess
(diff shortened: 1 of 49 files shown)


emeryberger:
Hi - we already use `black` (though it should be integrated into CI). I don't want to make any changes to a number of integrated bits of code - just actual Scalene code. If you have some concrete code improvements (not formatting), we would welcome those as broken out into a separate PR. Thanks!

PR #480, opened 2022-11-08
Poisson sampler

emeryberger:


emeryberger:
Merging in a Poisson (actually geometric distribution based) sampler, for experimental comparison only.

PR #484, opened 2022-11-09
Improve multiprocessing and module support

### LINKED ISSUE #126: Why is sys.executable wrapped in os.path.basename?
Hello,

I have just installed scalene 1.1.15 from PyPI using a virtualenv-based Python binary.

I had some difficulties with getting it to run and I pinned it down to the "os.path.basename(sys.executable)" line in the snippet below. For reference, the error message was "/usr/bin/python: No module named scalene".

I believe the assumption here is that the command "python" will always be the one that has scalene on its PYTHONPATH.

However, if my Python installation is in /home/user/projects/myproject/env/bin/python then using basename will trim it to "python" which points to the system-wide Python installation where scalene cannot be found.

I wonder why basename was used in this location? Perhaps it is important to keep it there and I just do not understand it yet, being a new user.

Thank you.

```python
    # Load the shared object on Linux.
    if sys.platform == "linux":
        if ("LD_PRELOAD" not in os.environ) and (
            "PYTHONMALLOC" not in os.environ
        ):
            os.environ["LD_PRELOAD"] = os.path.join(
                os.path.dirname(__file__), "libscalene.so"
            )
            os.environ["PYTHONMALLOC"] = "malloc"
            args = sys.argv[1:]
            args = [
                os.path.basename(sys.executable),
                "-m",
                "scalene",
            ] + args
```

### LINKED ISSUE #244: Profiling python module

### LINKED ISSUE #391: Running scalene when running python in modular mode
Hi,

I run my python code in the following way `python -m HumanObj.train --configs_yml='configs/v1.yml'`.  If I do not use the `-m` option python will not find my code. If I want to use `scalene` to run my code I get the following error:
`Scalene: could not find input file -m`. 
What do I need to do to get my code running and profile it?

```
python -m scalene HumanObj.train --configs_yml='configs/v1.yml'
Scalene: could not find input file HumanObj.train
```
However, this `python -m HumanObj.train --configs_yml='configs/v1.yml'` works totally fine.

Thanks,

RuRo:
Fixes #126

By setting `sys.executable` to an absolute path.

---

Fixes #391 
Fixes #244

By replacing `-m some.module` by the path to this module (eg `/usr/lib/python3.10/site-packages/some/module.py` or `/usr/lib/python3.10/site-packages/some/module/__main__.py`).

---

Let me know if you have any suggestions/questions.

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -677,7 +677,10 @@ class Scalene:
                 + os.environ["PATH"]
             )
             # Force the executable (if anyone invokes it later) to point to one of our aliases.
-            sys.executable = Scalene.__all_python_names[0]
+            sys.executable = os.path.join(
+                Scalene.__python_alias_dir,
+                Scalene.__all_python_names[0],
+            )
 
         # Register the exit handler to run when the program terminates or we quit.
         atexit.register(Scalene.exit_handler)
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -20,6 +20,7 @@ import contextlib
 import functools
 import gc
 import http.server
+import importlib.util
 import inspect
 import json
 import math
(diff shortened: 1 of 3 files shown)


emeryberger:
Great, many thanks!

PR #511, opened 2022-12-10
Add PEP 517 support (closes #503)

### LINKED ISSUE #503: Support PEP 517/518
**Is your feature request related to a problem? Please describe.**
Not directly, but I am filing this issue after a short chat on the Boston Python Slack about how to update one of the project's dependencies.

**Describe the solution you'd like**
To satisfy this request, `scalene` should support build systems that follow [PEP 517](https://peps.python.org/pep-0517), which mostly comes down to following [PEP 518](https://peps.python.org/pep-0518/) by introducing a `pyproject.toml` file to the project's source tree.

The biggest advantage of this approach is that build-time dependencies (namely `Cython` and possibly `crdp` as well) can be declared as such in the modern style, so end-user installation should be slightly smoother.

**Describe alternatives you've considered**
N/A

**Additional context**
I sat down for an hour or so tonight and saw how much of the project's `setup.py` could be migrated to a `pyproject.toml`. The biggest snag I ran into was the `DEV_BUILD` functionality that appends a `.devN` suffix to builds. And of course all the dynamic stuff related to available compilers and extension modules still needs to be expressed in `setup.py`

The upshot here is that it's pretty easy to start small with a `pyproject.toml`, it is _not_ a wholesale replacement for `setup.py` (or its declarative counterpart `setup.cfg`), as described [in the setuptools documentation](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html).

SnoopJ:
**NOTE: This changeset should be merged with care because of the changes it makes to `scalene`'s build workflows. At a minimum it should be tested on MacOS and Windows before merging!**

This PR introduces a `pyproject.toml` to bring `scalene` into the world of PEP 517/518. `setuptools` still does most of the lifting, but most of the configuration is now declared in this more modern file. One big advantage of this change is that build-time dependencies are clearly separated from runtime dependencies, so end-users should not need to worry about installing packages like `Cython` or `wheel` in order to build the project. I _think_ this means `requirements.txt` can be removed from the project, as it is no longer necessary in the GitHub workflow that depends on it, but I would appreciate maintainer insight on this.

### Pre-merge checklist

This checklist may be incomplete, please add to it if anything is missing.

* [x] Test build/install on Linux
  * I can install (with `pip install .`) and build (with `make bdist` and `make sdist`) with a fresh 3.8.12 venv on Ubuntu 20.04.3
  * I also checked that `DEV_BUILD` is respected when performing a build, adding a `.devN` tag to the resulting package version
* [ ] Test build/install on MacOS
  * The MacOS `smoketests` passed, so `pip install` looks good. It would be a good idea to check the Makefile-driven workflow too (unfortunately, I don't have a Mac to test with)
* [ ] Test build/install on Windows
  * I was able to `pip install` the repository on Windows 10 in a fresh Python 3.8.10 environment, but the Makefile-driven workflow should be checked too


pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+]
+dependencies = [
+    "cloudpickle>=1.5.0",
+    "ipython==7.31.1",
+    "Jinja2>=3.0.3",
+    "rich>=10.7.0",
+    "pynvml>=11.0.0",
+    "pyperf==2.0.0",
+    "lxml==4.9.1",
+    "packaging==20.9",
```
SnoopJ, line 47: I included these dependencies from the original `requirements.txt` but I don't see any usage of them by the project, and it builds without them, they can probably be removed.

```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+    "License :: OSI Approved :: Apache Software License",
+    "Operating System :: POSIX :: Linux",
+    "Operating System :: MacOS :: MacOS X",
+    "Operating System :: Microsoft :: Windows",
+]
+readme = "README.md"
+authors = [
+    {name = "Emery Berger", email = "emery@cs.umass.edu"},
+    {name = "Sam Stern"},
+    {name = "Juan Altmayer Pizzorno"},
```
SnoopJ, line 37: Another nice feature of `pyproject.toml` is that it makes it easy to have multiple authors. I've taken the liberty of adding the other authors listed in the project's README :sweat_smile:

pyproject.toml
```diff
@@ -0,0 +1,57 @@
+[build-system]
+requires = ["setuptools>=61.0.0", "setuptools_scm", "wheel>=0.36.1", "Cython>=0.29.28"]
+build-backend = "setuptools.build_meta"
```
SnoopJ, line 3: As far as I can tell, `setuptools_scm` is not actually used by the project, and it builds without them. It probably be removed from these dependencies.

```suggestion
requires = ["setuptools>=61.0.0", "wheel>=0.36.1", "Cython>=0.29.28"]
build-backend = "setuptools.build_meta"
```

jaltmayerpizzorno:
@SnoopJ I've re-targeted the PR to a branch I just opened for this (pep_517, original, right?). I'll work on it more in a few.

pyproject.toml
jaltmayerpizzorno: I don't see it used, either.

SnoopJ:
> @SnoopJ I've re-targeted the PR to a branch I just opened for this (pep_517, original, right?). I'll work on it more in a few.

Sounds good, let me know if you have any questions or requests :grin:

jaltmayerpizzorno:
I'm going on a short trip, so it seems best to postpone working on this until I'm back (in the 2nd week of January).

PR #552, opened 2023-01-20
Collect stack traces (not yet exposed in the AI)

emeryberger:
Includes some code cleanup, modification of the prompt, and tweaks to the UI.


scalene/hashablelist.py
```diff
@@ -7,49 +7,64 @@ class HashableList(list):
...
         self._hash = None
 
-    def extend(self, iterable):
-        super().extend(iterable)
-        self._hash = None
+    def append(self, item) -> None:
+        # Ensure we have a hash value.
+        self.hash()
+        super().append(item)
+        self._hash.update(str(hash(item)).encode())
```
janne-tervo, line 51: This optimization is incompatible with self._hash being an integer (self._calculate_hash returns int).

self._hash can't be HASH either, because openssl hashes are not serializable.

Same bug is repeated below on new function extend() in scalene/hashatablelist.py:47.

The previous version of this function seems to works.

PR #582, opened 2023-03-18
Moving to GPT-4.

emeryberger:
Made changes so Scalene works with GPT-3.5 instead of GPT-3. Initial "pure Python" prompt added, not yet exposed in the API.


17Reset:
Why don't you make an option on the ui to choose GPT-3.5 or GPT-4? Because some people don't have a key for GPT-4.

Issue #587, opened 2023-03-28
is_native doesn't handle the TypeError case from os.path.dirname.


ahaberlach:
**Describe the bug**

"is_native" check seems to be failing to catch `TypeError` which is returned by `os.path.dirname(None)`

I was running into this while profiling some native code, and it may be a python 3.8 or older system issue.

```
"~/.virtualenv/production/lib/python3.8/site-packages/scalene/scalene_analysis.py" 137L, 5125C                                                                          23,13         Top
  File "/home/ubuntu/.virtualenv/production/lib/python3.8/site-packages/scalene/scalene_profiler.py", line 932, in output_profile
    json_output = Scalene.__json.output_profiles(
  File "/home/ubuntu/.virtualenv/production/lib/python3.8/site-packages/scalene/scalene_json.py", line 393, in output_profiles
    imports = ScaleneAnalysis.get_native_imported_modules(code_str)
  File "/home/ubuntu/.virtualenv/production/lib/python3.8/site-packages/scalene/scalene_analysis.py", line 93, in get_native_imported_modules
    if ScaleneAnalysis.is_native(node.module):
  File "/home/ubuntu/.virtualenv/production/lib/python3.8/site-packages/scalene/scalene_analysis.py", line 23, in is_native
    package_dir = os.path.dirname(package.__file__)
  File "/usr/lib/python3.8/posixpath.py", line 152, in dirname
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

I fixed/worked around by adding the check for TypeError to scalene_analysis.py.  I've got a PR on the way.

**To Reproduce**

Please include a minimum working example if at all possible.

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

If you have not yet tried with the repository version (`python3 -m pip install git+https://github.com/plasma-umass/scalene`), please try that before reporting.

**Additional context**

ahaberlach:
https://github.com/plasma-umass/scalene/pull/588 should fix this -- anybody know who can review?

emeryberger:
I believe this is fixed. Please let me know either way!

PR #588, opened 2023-03-28
Add handling of TypeError case to is_native

ahaberlach:
This should address https://github.com/plasma-umass/scalene/issues/587 which has more information, but basically the check for native fails on at least my system because package.__file__ exists, but is None, so os.path.dirname() returns a TypeError.

### DIFF
--- a/scalene/scalene_analysis.py
+++ b/scalene/scalene_analysis.py
@@ -31,6 +31,9 @@ class ScaleneAnalysis:
         except AttributeError:
             # No __file__, meaning it's built-in. Let's call it native.
             result = True
+        except TypeError:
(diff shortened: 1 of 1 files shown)


emeryberger:
Can you print out some info so we can see what library this is? It's a surprising issue. Thanks!

akx:
Saw #587 happen to me locally just now and was about to come report it, but here we go :)

The `package` for me is

```
<module 'google.cloud' (<_frozen_importlib_external.NamespaceLoader object at 0x103b68060>)>
```

emeryberger:
This should be fixed in the latest release!

PR #620, opened 2023-06-04
Switch from per object headers to out-of-band hash map

emeryberger:
We currently prepend headers to each object that contain the object size (see https://github.com/plasma-umass/scalene/blob/master/src/include/scaleneheader.hpp). We do this because small objects are allocated by Python so we can't use `malloc_usable_size` (or similar) to get the size. (Note that we need the sizes so we can properly account for memory consumption, particularly when we free objects.)

Unfortunately, there are cases when interposition is not early enough to catch all calls to `malloc`, and in some cases, those objects are then passed to `free` or `realloc`. Such objects have no headers. If offsetting to before the object (where the header would be) happens to land in unmapped memory, game over (segfault).

This PR addresses this issue, adding a `SizeManager` that manages all object sizes out of band in a concurrent hash map. In the case outlined above, instead of segfaulting, will return a size of 0 for the uninterposed object.

### DIFF
--- a/src/include/mmapallocator.hpp
+++ b/src/include/mmapallocator.hpp
@@ -1,0 +1,25 @@
+#pragma once
+
+#include <sys/mman.h>
+
+template <typename T>
+class MmapAllocator {
+public:
+  using value_type = T;
+  
+  MmapAllocator() {
+  }
+  
+  template <typename U>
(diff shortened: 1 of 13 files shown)


emeryberger:
Closing this PR because it yields unacceptably high overhead, and I also can't reproduce the problem it addressed.

PR #637, opened 2023-07-08
Updates to properly handle @profile decorators for CPU time attribution

emeryberger:
Fixes to restore the `@profile` decorator to its intended semantics (if any code has a `@profile` decorator, only profile that code).

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -417,9 +417,6 @@ class Scalene:
             # Different line: stop tracing this frame.
             frame.f_trace = None
             frame.f_trace_lines = False
-            # If we are not in a file we should be tracing, return.
-            # if not Scalene.should_trace(ff):
-            #     return None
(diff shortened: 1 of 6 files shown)


emeryberger:
@sternj We could use a test for `@profile` to avoid future regressions.

scalene/scalene_profiler.py
```diff
@@ -1595,13 +1605,18 @@ def memcpy_sigqueue_processor(
 
     @staticmethod
     @functools.lru_cache(None)
-    def should_trace(filename: str) -> bool:
+    def should_trace(filename: str, func : str) -> bool:
```
emeryberger, line 1608: Fix documentation line below (if file name *and* function)…

emeryberger: (I forgot to update it)

emeryberger:
Smoketest is currently failing…

Issue #632, opened 2023-06-29
Signal Scalene to report what it has gathered


garethky:
**Is your feature request related to a problem? Please describe.**
I'm trying to use Scalene to profile a large, multi threaded, Python program that's designed to be a system service. The only way the service exits is via the shutdown command. There is no internal way to cleanly exit the service. When testing on the command like I have to kill the process to make it exit.

Basically I need a profiling report from a chunk of runtime of a program without the program having to exit.

**Describe the solution you'd like**
I'd like a way to signal to Scalene that it should prepare the report based on the data it has already gathered and write it to the output file. Something like:

```
$ scalene --profile-all --outfile ~/profile.json service.py &
$ python3 -m scalene.profile --off --pid 1337
$ echo "10 minutes later..."
$ python3 -m scalene.profile --off --pid 1337
$ python3 -m scalene.profile --report --pid 1337
```
and the report shows up at `~/profile.json` immediately.

If I later toggle the profiling and ask for a report, it will be a new report for the new chunk of runtime that was profiled.

**Describe alternatives you've considered**
The only way I can get output from this program is with `--profile-interval`

**Additional context**
The service in question is the open source [klipper 3D printer 'firmware'](https://www.klipper3d.org/), which is mainly written in Python.

I think other people have tried to ask for this but they didn't make clear what they needed. Some have asked to connect Scalene to a running process. This is a way to get a report without having to capture the entire run of the program or wait for it to exit to see results.

emeryberger:
Hi @garethky - thanks for your note. There are not many signals available that we haven't already hijacked, so I'm leery of grabbing another one. That said, what if `--off` always produced a new profile? I'm trying to think of a use case for `--off` that would be compromised by that update. Another option is to make `--off` and `--on` actually output the same signal, but just toggle the behavior, and then add a `--report` or similar.

garethky:
I would be fine with `--off` causing the output to be written to. That makes a lot of sense to me.

Is there any problem in the case where an `--outfile` isn't specified and you need to write to stdout?

emeryberger:
I am thinking it would make sense to suppress the output for that case, but am not certain. Thoughts?

emeryberger:
Prototype of functionality here: `python3 -m pip install --force-reinstall git+https://github.com/plasma-umass/scalene@off_report`

garethky:
Sorry, I haven't forgotten about this, just got very busy. What you are proposing looks good to me.

[...]

PR #641, opened 2023-07-24
Force profile output when triggered by --off.

emeryberger:


emeryberger:
Discussion here: https://github.com/plasma-umass/scalene/issues/632

PR #646, opened 2023-07-29
Add  when invoking with -m

### LINKED ISSUE #539: Profiling using `-m mypackage` fails if the module's `__main__.py` contains relative imports
**Describe the bug**
Support for profiling modules that would ordinarily be run via the `-m mypackage` syntax (which uses a package's `__main__.py` entry point) was added recently in #484, fixing #244.

However, if this feature is used with a package whose `__main__.py` file contains relative imports, then profiling fails with `ModuleNotFoundError`.

I believe this is due to the approach used to load modules - which is to convert the `-m mypackage` argument to instead reference the `__main__.py` file directly - which bypasses the usual module loading process.

**To Reproduce**
1. `mkdir testcase && cd $_`
2. `python -m venv .venv && source .venv/bin/activate`
3. `pip install scalene==1.5.18 salesforce-functions==0.2.0`
4. `scalene -m salesforce_functions`

**Expected behavior**

Scalene successfully runs. (With the reduced example above, a profile won't actually be created since the program doesn't run for long enough, but Scalene itself shouldn't error.)

Or as a last resort, Scalene should gracefully handle the `ModuleNotFoundError` and warn that relative imports in `__main__.py` aren't currently supported.

**Actual behavior**

```
(.venv) testcase $ scalene -m salesforce_functions
 No module named 'scalene._internal'
  File "/Users/emorley/src/testcase/.venv/lib/python3.11/site-packages/scalene/scalene_profiler.py", line 1784, in profile_code
  File "/Users/emorley/src/testcase/.venv/lib/python3.11/site-packages/salesforce_functions/__main__.py", line 3, in <module>
    from ._internal import cli
ModuleNotFoundError: No module named 'scalene._internal'
Scalene: Program did not run for long enough to profile.
```

The failing line corresponds to:
https://github.com/heroku/sf-functions-python/blob/v0.2.0/salesforce_functions/__main__.py#L3

Note the `scalene._internal` part of the error message, which comes from the `__main__.py` file's `from ._internal import cli` line being incorrectly treated as relative to the `scalene` package, not the `salesforce_functions` package.

If I change that line to be an absolute import (ie: `from salesforce_functions._internal import cli`), then Scalene works. (Relative imports can be left as is in the rest of the `salesforce_functions` project, just not `__main__.py`)

**Additional context**
 - macOS 12.6 Monterey, on an M1 MacBook Pro
 - Homebrew Python 3.11.1
 - The error occurs with both Scalene 1.5.18 (per the steps to reproduce above), and also if I install from `main` of the GitHub repository.
 - The error occurs with both the `scalene` and `python -m scalene` CLI syntax styles.

sternj:
Fixes #539

### DIFF
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -2071,14 +2071,17 @@ class Scalene:
                     the_globals = __main__.__dict__
                     # Splice in the name of the file being executed instead of the profiler.
                     the_globals["__file__"] = os.path.abspath(progs[0])
-                    # Some mysterious module foo to make this work the same with -m as with `scalene`.
-                    # the_globals["__spec__"] = spec
-                    # print("NAME", spec.name)
+                    # This part works because of the order in which Python attempts to resolve names--
+                    # Within a given context, it first tries to look for __package__, and then for __spec__.
+                    # __spec__ is a ModuleSpec object that carries a lot of extra machinery and requires
+                    # extra effort to create (it seems, at least).
+                    #
+                    # __spec__ was originally set to none because the __globals__ here has the Scalene ModuleSpec
+                    # but it doesn't seem like that was enough. Setting the __package__, as below, seems to be enough to make
+                    # it look in the right place
+                    the_globals["__spec__"] = None 
                     if spec is not None:
-                        name = spec.name
-                        # MAIN_MOD = '.__name__'
-                        # if name.endswith(MAIN_MOD):
-                            # name = name[:-len(MAIN_MOD)]
+                        name = spec.name 
                         the_globals['__package__'] = name.split('.')[0]
                     # Do a GC before we start.
                     gc.collect()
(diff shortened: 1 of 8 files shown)


scalene/scalene_profiler.py
```diff
@@ -800,6 +800,10 @@ def __init__(
                 os.chmod(fname, stat.S_IXUSR | stat.S_IRUSR | stat.S_IWUSR)
             # Finally, insert this directory into the path.
             sys.path.insert(0, str(Scalene.__python_alias_dir))
+            # print("PACKJAGE", __package__)w
```
emeryberger, line 803: Remove comment cruft

scalene/scalene_profiler.py
```diff
@@ -2070,7 +2072,14 @@ def run_profiler(
                     # Splice in the name of the file being executed instead of the profiler.
                     the_globals["__file__"] = os.path.abspath(progs[0])
                     # Some mysterious module foo to make this work the same with -m as with `scalene`.
-                    the_globals["__spec__"] = None
+                    # the_globals["__spec__"] = spec
```
emeryberger, line 2075: Remove comment cruft - I'm not thrilled with "mysterious module foo" as a comment :). Can you point to any resource here, or elaborate on what's going on?

PR #648, opened 2023-08-04
Remove vsnprintf definition from printf.h

leavelet:
This definition will break STL string, starting from GCC13.

Although GCC13 will not being widely used soon, we have encountered such problem on Arch Linux for loong64.


leavelet:
By the way, I am wondering why we should use the `printf` library, which is not designed for C++.

I think printf from stdio will work very well in our case. Could anyone please kindly tell me the reason?

PR #658, opened 2023-08-18
fix handling of `force_print` in scalene_json.py

### LINKED ISSUE #657: `force_print` argument for output behaves strange
https://github.com/plasma-umass/scalene/blob/6980cb0c41df6ae724a2ce8d4686398b4b058853/scalene/scalene_json.py#L109C12-L109C15

Shouldn't it be `if force_print and not ...`? The printing will happen if `force_print=False`.

danlkv:
Fixes #657


sternj:
Currently we do rely on the JSON output containing a record of every line for indexing purposes in the UI. I think if you want to filter things out of the JSON, you can programmatically do that on an ad-hoc basis

PR #667, opened 2023-08-23
Added invocation arguments to json

sternj:
Adds an extra field to Scalene JSON with the invocation arguments


emeryberger:
CodeQL is correctly failing.

PR #680, opened 2023-09-03
Simplify __functions_to_profile construction.

ttsugriy:
Using lambda only adds overhead and is less idiomatic.


emeryberger:
Thanks!

PR #685, opened 2023-09-04
Remove unnecessary list creation in scalene_analysis.py

ttsugriy:
`join` takes an iterable, so passing the result of `map` works and doesn't incur extra memory and CPU cost required to construct `list`.


emeryberger:
Would you mind combining all of these into a single omnibus PR? Thanks!

ttsugriy:
sorry @emeryberger but I don't have a local checkout and sent these PRs using github web editor that doesn't allow merging.

PR #688, opened 2023-09-06
Added function name to `stack`

sternj:


scalene/scalene_profiler.py
```diff
@@ -929,17 +929,19 @@ def flamegraph_format() -> str:
...
             output += "\n"
         return output
 
     @staticmethod
     def output_profile(program_args: Optional[List[str]] = None) -> bool:
+        """Output the profile. Returns true iff there was any info reported the profile."""
         # sourcery skip: inline-immediately-returned-variable
         # print(Scalene.flamegraph_format())
-        """Output the profile. Returns true iff there was any info reported the profile."""
+        with open(f'{os.getcwd()}/flamegraph.out', 'w+') as f:
```
emeryberger, line 943: This should not be output by default.

emeryberger (changes requested):

PR #705, opened 2023-09-22
No check for browser if --no-browser

gb03db:


emeryberger:
Seems innocuous but can you provide some motivation for this? Thanks!

emeryberger:
This feature has been added.

PR #712, opened 2023-10-11
Send and receive thread-level memory signals.

emeryberger:
WIP. Addresses memory accounting issue identified here: https://github.com/plasma-umass/scalene/discussions/659.


emeryberger:
This has been obsoleted.

Issue #717, opened 2023-10-17
Display percentage and absolute time *values* in the html report


dhirschfeld:
**Is your feature request related to a problem? Please describe.**

The bar charts showing the percentage of time on each line are useful to visually identify and quickly hone in on the problematic lines.

Once you've identified the problematic lines you want to get details such as the percentage time and the absolute time spent on each line in the function. This information is missing from the HTML report. The percentage time you can get from hovering over the bar chart, but that's extremely cumbersome to do for each line and doesn't let you see all the data at once. Also I would like to see the absolute times - for me, they're more important than the percentages, but both are useful to have.

**Describe the solution you'd like**
Show `time %` and `time (ms)` as columns in the HTML report next to each line.

ioannis-vm:
This is a much needed feature. It seems straightforward to have a command line flag that controls this behavior, chosing between percentages or valusees.

mucmch:
Fully agree with @dhirschfeld.
Having `time %` and `time (ms)` as extra columns would be very helpful.
Currently, this information is not available at all for me as hovering over does not work.

emeryberger:
Time is now visible when hovering over the graphical time bars.

PR #718, opened 2023-10-22
Add numbers to CPU and memory bars.

emeryberger:
This PR adds numbers to bars to avoid the need to mouse over entries (though that tooltip functionality remains). To ensure that labels fit, numbers only appear on segments comprising at least 20% of the total. The screenshot below presents an example profile with the new functionality. Partially addresses https://github.com/plasma-umass/scalene/issues/717.

Screenshot 2023-10-22 at 10 26 50 AM

### DIFF
--- a/scalene/scalene-gui/scalene-gui.js
+++ b/scalene/scalene-gui/scalene-gui.js
@@ -323,15 +323,16 @@ async function copyOnClick(event, message) {
 }
 
 function memory_consumed_str(size_in_mb) {
+    console.log(`size in mb = ${size_in_mb}`);
   // Return a string corresponding to amount of memory consumed.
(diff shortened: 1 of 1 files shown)


ioannis-vm:
It's great that we have those numbers there, but now the hover popup is gone. Is that the case on your end as well?

```
Scalene version: 1.5.38 installed from PyPI.
Python: 3.10.13
System: Arch Linux, 6.8.1-arch1-1 kernel.
```

emeryberger:
I couldn't figure out how to place the hover popup so it was still readable. I tried but gave up. Clicking on the CPU bars to flip between percentage and "time elapsed" seems doable - please open a separate issue (this PR is now out of date, so I am closing it).

Issue #723, opened 2023-11-08
Handle browsers not in cpython's webbrowser.py


Jeremiah-England:
**Describe the bug**

Browsers that aren't explicitly handled in [`webbrowser.py`](https://github.com/python/cpython/blob/main/Lib/webbrowser.py) end up as `GenericBrowser` objects when the BROWSER environment variable is set. Scalene won't open the web UI for those browsers because we try to avoid opening in text browsers and it seemed like `GenericBrowser` was for those.

https://github.com/plasma-umass/scalene/blob/master/scalene/scalene_profiler.py#L1768-L1775

This is probably why @munael's [comment](https://github.com/plasma-umass/scalene/issues/383#issuecomment-1299123915) has upvotes.

**To Reproduce**

1. Set the BROWSER environment variable to vivaldi (or brave probably too).
2. Run `scalene --viewer`

**Desktop:**
 - OS: Archlinux
 - Browser: Vivaldi
 - Version: 1.5.31.1

**Additional context**

Lines in webbrowser.py that create the `GenericBrowser`: https://github.com/python/cpython/blob/main/Lib/webbrowser.py#L528-L540

When I remove the BROWSER environment variable, I get a `BackgroundBrowser` object with a name of `xdg-open`.

I'm happy to write up a patch for this if you want. I think we should probably just explicitly list out the common text browsers and ignore them. We could start with the list from `webbrowser.py` [here](https://github.com/python/cpython/blob/main/Lib/webbrowser.py#L511-L526).

Jeremiah-England:
@emeryberger, I submitted a PR (#724). Let me know if you see anything I should change!

Also, thanks for working on scalene!

Jeremiah-England:
Thanks for the merge @emeryberger!

I'm going to close this now. Feel free to re-open if you like to leave things open till an actual release.

emeryberger:
Thanks for the contribution!

PR #724, opened 2023-11-09
Handle text browsers explicitly to support unlisted GUI browsers

Jeremiah-England:
Ref: https://github.com/plasma-umass/scalene/issues/723

I tested by:

- Setting BROWSER to lynx and making sure it still defaulted to printing to the console instead of opening the html file.
- Setting BROWSER to vivaldi and making sure it did open the web view by default.
- Setting BROWSER to chromium and making sure it still opened the web view by default.
- Passing --cli with BROWSER=vivaldi to make sure it just printed to the console.
- Installing jupyter-lab to the venv, opening it, and running `%scrun` and `%%scalene` on some cells. I got the iframe to render with profile results.

### DIFF
--- a/scalene/find_browser.py
+++ b/scalene/find_browser.py
@@ -1,0 +1,12 @@
+def find_browser():
+    """Find the default browser if possible and if compatible."""
+    import webbrowser
+
+    # Mostly taken from the standard library webbrowser module. Search "console browsers" in there.
+    # In general, a browser belongs on this list of the scalene web GUI doesn't work in it.
+    # See https://github.com/plasma-umass/scalene/issues/723.
+    incompatible_browsers = {"www-browser", "links", "elinks", "lynx", "w3m", "links2", "links-g"}
+
+    browser = webbrowser.get()
+
+    return None if browser.name in incompatible_browsers else browser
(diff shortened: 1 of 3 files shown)


scalene/find_browser.py
```diff
@@ -0,0 +1,12 @@
+def find_browser():
+    """Find the default browser if possible and if compatible."""
+    import webbrowser
+
+    # Mostly taken from the standard library webbrowser module. Search "console browsers" in there.
+    # In general, a browser belongs on this list of the scalene web GUI doesn't work in it.
+    # See https://github.com/plasma-umass/scalene/issues/723.
+    incompatible_browsers = {"www-browser", "links", "elinks", "lynx", "w3m", "links2", "links-g"}
+
+    browser = webbrowser.get()
```
Jeremiah-England, line 10: One thing to notice here is that we are no longer checking if `webbrowser.get()` is truthy. Looking at the code, it doesn't look like it will ever return None. It looks like it will raise `webbrowser.Error` if one cannot be found. If we want to handle that case I think we will need a try/catch.

Another reason besides reading the code itself to think that it cannot return `None` is that is how typeshed has typed it here: https://github.com/python/typeshed/blob/main/stdlib/webbrowser.pyi#L13C38-L13C38

emeryberger: Agreed on the truthiness, thanks. How about degrading gracefully (in the face of an exception) to use the CLI?

Jeremiah-England: @emeryberger, good idea. I added a try/except in there.

PR #750, opened 2023-12-30
Initial native stack collection (works only with local fork of Python).

emeryberger:
This PR adds preliminary support (not exposed in the UI) for native stack collection. It depends on this fork that adds support for native stack collection to the CPython interpreter: https://github.com/emeryberger/cpython.


emeryberger:
Closing — this approach requires a custom CPython fork, which isn't a viable path. Going to pursue external stack unwinding (libunwind / .eh_frame, à la py-spy/memray) instead, which doesn't require upstream interpreter changes.

PR #757, opened 2024-01-18
ci: re-enable py3.12

purificant:
This PR re-enables testing with Python 3.12 and addresses the [issue](https://github.com/plasma-umass/scalene/actions/runs/7561817150/job/20591132945) that was preventing CI from passing before.

Notable changes:
`lxml` updated to latest version to enable compatibility with py3.12
`pip` version lock is removed because we are using the latest version in CI through `python -m pip install --upgrade pip`


emeryberger:
Thanks, much appreciated!

PR #776, opened 2024-02-18
Disable GCC's "unused-result" warnings

sarahec:
GCC fails when building as it doesn't like the printf redefinition using ::write. Write apprently returns a value, even if nobody really wants it.

The addition of unused result checking caused all kinds of griping in the GCC userbase. This turns that check off globally. (Note, with the check on, the only failing line is the printf redefinition).


emeryberger:
Thanks! Fixes https://github.com/plasma-umass/scalene/issues/777.

PR #779, opened 2024-02-19
Always use system browser

### LINKED ISSUE #778: Web browser doesn't launch on NixOS
**Describe the bug**
When I run scalene, I only get the console output; the web page never opens. 

(*Note:* PR incoming)

 - OS: NixOS Linux (6.1.77 kernel, on the unstable channel)
 - Browser Chrome
 - Version [e.g. 22]  121.0.6167.184

**Additional context**
NixOS is an "immutable" distribution where the system binaries are all read-only. It's used widely in the financial industry (who need to only use security-vetted tools and libraries) and has a huge following. (More on immutable distros: https://itsfoss.com/immutable-linux-distros/ )

Immutable systems typically put their binaries in a special directory and use symlinks or other means to reference the binaries. So looking for "google-chrome" or "google-chrome-stable" won't work. But the Python `webbrowser` module knows how to find and launch it. 

Also, the scalene code does quite a bit of filtering to find "suitable" browsers. I suggest we just use the system browser and can either 1) document compatibility (i.e. any reasonably modern browser), or 2) include a capability check in the webpage.

sarahec:
Simplifies the code by removing logic to pick from a set of known browsers. Launches using the system browser.

If this becomes a problem, we can always ask the user to supply a suitable one via the command line or print the launch URL to the console and have them click it. 

Fixes #778.

### DIFF
--- a/scalene/find_browser.py
+++ b/scalene/find_browser.py
@@ -1,13 +1,18 @@
 import webbrowser
 from typing import Optional
 
-
 def find_browser(browserClass=None) -> Optional[str]:
-    """Find the default system browser"""
+    """Find the default system browser, excluding text browsers.
+    
+    If you want a specific browser, pass its class as an argument."""
+    text_browsers = [
+        "browsh", "elinks", "lynx", "w3m",
+    ]
+
     try:
         # Get the default browser object
         browser = webbrowser.get(browserClass)
-        return browser.name
+        return browser.name if browser.name not in text_browsers else None
     except webbrowser.Error:
         # Return None if there is an error in getting the browser
         return None
--- a/scalene/find_browser.py
+++ b/scalene/find_browser.py
@@ -6,7 +6,7 @@ def find_browser(browserClass=None) -> Optional[str]:
     
     If you want a specific browser, pass its class as an argument."""
     text_browsers = [
-        "browsh", "elinks", "lynx", "w3m",
+        "browsh", "elinks", "links", "lynx", "w3m",
     ]
 
     try:
--- a/scalene/find_browser.py
+++ b/scalene/find_browser.py
@@ -2,34 +2,12 @@ import webbrowser
 from typing import Optional
 
 
-def find_browser() -> Optional[webbrowser.BaseBrowser]:
-    """Find the default browser if possible and if compatible."""
-    # Names of known graphical browsers as per Python's webbrowser documentation
-    graphical_browsers = [
-        "windowsdefault",
-        "macosx",
-        "safari",
-        "google-chrome",
-        "chrome",
-        "chromium",
-        "firefox",
-        "opera",
-        "edge",
-        "mozilla",
-        "netscape",
-    ]
+def find_browser(browserClass=None) -> Optional[str]:
+    """Find the default system browser"""
     try:
         # Get the default browser object
-        browser = webbrowser.get()
-        # Check if the browser's class name matches any of the known graphical browsers
-        browser_class_name = str(type(browser)).lower()
-        if any(
-            graphical_browser in browser_class_name
-            for graphical_browser in graphical_browsers
-        ):
-            return browser
-        else:
-            return None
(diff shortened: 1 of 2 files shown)


emeryberger:
The logic that is in there was there for two reasons: (1) allowing us to launch an insecure Chrome browser that can open local file URLs; I eschewed that approach in favor of launching a local server; (2) getting the "right" behavior when running on headless systems that have something like `lynx` installed as the system browser, which definitely raises havoc. While we could remove the old insecure launch code, it took some time to figure out how to get that working, which is why it is still in the repo. The non-graphical browser problem, though, needs to be addressed.

sarahec:
One more thought: you were absolutely right to have a web server. 

What if the fix for headless servers is to write the page and test results (with unique names) to an external directory where they can be served via a web server? You can pass the result file name in a url parameter.

This would enable running batches of tests and reviewing them at leisure. It also opens the door to comparing the results of changes.

(Sorry if this is a lot of changes at once! Apparently I haven't lost my Google work habits.)

scalene/find_browser.py
```diff
@@ -1,35 +1,18 @@
...
-        "opera",
-        "edge",
-        "mozilla",
-        "netscape",
+def find_browser(browserClass=None) -> Optional[str]:
+    """Find the default system browser, excluding text browsers.
+    
+    If you want a specific browser, pass its class as an argument."""
+    text_browsers = [
+        "browsh", "elinks", "lynx", "w3m",
```
emeryberger, line 9: Missing `links` browser, mentioned [here](https://docs.python.org/3/library/webbrowser.html#webbrowser.register).

scalene/launchbrowser.py
```diff
@@ -21,31 +21,6 @@ def read_file_content(directory: str, subdirectory: str, filename: str) -> str:
     file_path = os.path.join(directory, subdirectory, filename)
     return pathlib.Path(file_path).read_text()
 
-
```
emeryberger, line 24: Restore just in case please :).

emeryberger (changes requested):

emeryberger:
Right now, the solution for headless servers is to output a JSON file (e.g., `--json --outfile foo.json`, or `--no-browser` which outputs `profile.json`) and load that file from `scalene --viewer`. I didn't quite follow the rest of your proposal, or at least how it would work in a common enough use case -- connection to some machine on Amazon or Azure or Google Cloud, where hosting a webserver is doable but maybe fraught?

sarahec:
Let me know if you want me to squash the changes

emeryberger:
Thanks!

sarahec (commented):
What if we move the propsal part to an open issue? To answer your question, let me unpack a bit:
1. A web server often has a static files directory (e.g. .html, .js,, .css).
2. The viewer page and its supporting files don't have to be assembled by scalene, but can live as static files. 
3. We still have to get the data into the viewer page, perhaps as a URL parameter. (Modern browsers also have a built-in file picker the page can use.)
4. On a headless server, the setup can include a web server that sees that directory. Locally, in Python, the built-in web server module also supports static directories.
5. Now the browser launch becomes `open("https://${host}:${port}/viewer.html?profile=${profile-path}")` . It works the same way locally or  on the development server. Locally could use a temp dir and fixed name by default.
6. You can support batch operation, etc. by writing the profile file as  `profile{$timestamp}.json` or, optionally,  `profile-$appname-$timestamp.json`. with the prefix coming from a command-line flag.
7. This assumes you have a default port number that can also be changed with a command-line flag.
8. If you're feeling wild and crazy, perhaps the app could use an optional configuration file.

Granted, it's a lot of little changes -- changing page generation, how the page gets its data, updating the server, documentation, etc. -- so it could qualify as a major release.  Having a series of runs available enables comparisons between them, taking multiple runs and averaging results, etc. 

This is all out of scope for the current PR! :)

Issue #784, opened 2024-02-25
AttributeError: 'str' object has no attribute 'suffix'


psykokwak-com:
**Describe the bug**
I just installed scalene with the following : 
`(venv) C:\Users\Jyce\Desktop\test>python -m pip install -U scalene`

Then execute : 
`(venv) C:\Users\Jyce\Desktop\test>python -m scalene run.py`

Here What I got : 

> Scalene failed to initialize.
> Traceback (most recent call last):
>   File "C:\Users\Jyce\Desktop\test\venv\lib\site-packages\scalene\scalene_profiler.py", line 2006, in run_profiler
>     profiler = Scalene(args, Filename(prog_name))
>   File "C:\Users\Jyce\Desktop\test\venv\lib\site-packages\scalene\scalene_profiler.py", line 707, in __init__
>     Scalene.__orig_python = redirect_python(preface, cmdline, Scalene.__python_alias_dir)
>   File "C:\Users\Jyce\Desktop\test\venv\lib\site-packages\scalene\redirect_python.py", line 47, in redirect_python
>     if sys.platform == "win32" and sys.executable.suffix == ".exe":
> AttributeError: 'str' object has no attribute 'suffix'

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - Python : 3.10.6

_If you have not yet tried with the repository version (`python3 -m pip install git+https://github.com/plasma-umass/scalene`), please try that before reporting._
=> Tried and I got the same error.

FSaal:
I am getting the same error using python 3.11.5.
Removing  str( from line 46 in redirect_python.py fixed this.
I don't know if redirect_python must return a str instead of a Path, if not, reverting the changes from commit #769 in redirect_python.exe would fix this issue.

LucasRr:
Getting the same issue here. Indeed, removing `str()` from l. 46 of `redirect_python.py` as FSaal suggested solved the issue.

emeryberger:
Thanks for the report. I'm waiting for the PR to go through CI. Once it does, I'll merge it and you should be able to verify that it works by installing from the repo (`python3 -m pip install git+https://github.com/plasma-umass/scalene`). Please report back!

emeryberger:
Tagging @alphabetagamer

alphabetagamer:
Ideally, no module should modify system variables, if we are using methods for a different type we can modify a copy of it. Updating the latest commit with this thought in mind, let me know if it aligns with the repository guidelines.

[...]

PR #789, opened 2024-03-05
Inconsistent behaviour between win & linux, this also clashes with lo…

alphabetagamer:
…ky spawn.py, loky backend spawn.py uses sys.executable expecting a str type

`
if sys.platform != 'win32':
    WINEXE = False
    WINSERVICE = False
else:
    import msvcrt
    from multiprocessing.reduction import duplicate
    WINEXE = (sys.platform == 'win32' and getattr(sys, 'frozen', False))
    WINSERVICE = str(sys.executable).lower().endswith("pythonservice.exe")
`


emeryberger:
Can you discuss this here esp. in light of the discussion and commits? https://github.com/plasma-umass/scalene/issues/784

emeryberger:
`sys.executable` is changed to support multiple process tracking so the test is probably correct. I'd recommend you look at that logic (which involves forcing Python processes to use Scalene instead of just invoking `python`).

emeryberger:
Looks like the required changes were previously implemented; closing.

PR #818, opened 2024-04-26
Smoketest fixes

sternj:
PR being created to run smoketests


emeryberger:
Is this still relevant? @sternj

sternj:
It is not. This was when we were having the "bad bind opcode" issue that @jaltmayerpizzorno and I were working on. This PR was experimenting with just not running on the problematic version because I couldn't reproduce locally, but Juan managed to and tracked down the actual root of the problem

PR #822, opened 2024-05-02
Move packaging to PEP517 style

jaltmayerpizzorno:
This PR adds a `pyproject.toml`, making our packaging PEP517 compatible.
I also cleaned up that kludgy way to set the platform for "-universal2" wheels on MacOS.

The MacOS 14 issues on GitHub remain, so I changed a couple of the matrix items for "tests" and "smoketests" to `macos-13` (so that I could test).  I didn't test the packaging separately on GitHub because the matrix there is more complex... I don't want to cause conflicts with anything that Sam may be working on for MacOS 14.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,15 @@
 from setuptools import setup, find_packages
 from setuptools.extension import Extension
-from scalene.scalene_config import scalene_version
 from os import path, environ
 import sys
 import sysconfig
+from pathlib import Path
+
+# needed for isolated environment
+sys.path.insert(0, str(Path(__file__).parent.resolve()))
+from scalene.scalene_config import scalene_version
+sys.path.pop(0)
+
 
 if sys.platform == 'darwin':
     import sysconfig
@@ -21,7 +27,6 @@ def compiler_archs(compiler: str):
     """Discovers what platforms the given compiler supports; intended for MacOS use"""
     import tempfile
     import subprocess
-    from pathlib import Path
 
     print(f"Compiler: {compiler}")
     arch_flags = []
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -160,75 +165,35 @@ def copy_extensions_to_source(self):
...
-    # which is hard to obtain (see BuildExtCommand above).
-    import platform as p
-    v = p.mac_ver()[0].split(".")
-    v = f"{v[0]}.0" if int(v[0]) >= 11 else ".".join(v)
-    sys.argv.extend(['--plat-name', f"macosx-{v}-universal2"])
+def bdist_wheel_options():
+    if sys.platform == 'darwin':
+        # Build universal wheels on MacOS.
+        # ---
+        # On MacOS >= 11, all builds are compatible for a major MacOS version, so Python "floors"
```
emeryberger, line 172: "compatible with"

setup.py
emeryberger, line 1: Can you explain the role of `setup.py` once we have a `pyproject.toml` in place? I assumed the latter would obviate the former but clearly not.

setup.py
```diff
@@ -1,9 +1,15 @@
 from setuptools import setup, find_packages
 from setuptools.extension import Extension
-from scalene.scalene_config import scalene_version
 from os import path, environ
 import sys
 import sysconfig
+from pathlib import Path
+
+# needed for isolated environment
```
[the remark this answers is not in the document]
jaltmayerpizzorno: The `import` of `scalene.scalene_config` fails within PEP517's build isolation environment unless we add the directory to the import path.

setup.py
jaltmayerpizzorno: Good catch :)  I made it "compatible within"

setup.py
jaltmayerpizzorno: You still need a [_build backend_](https://peps.python.org/pep-0517/#terminology-and-goals) to build the application. Some build backends don't support extension modules; ours still uses `setuptools` for that (which AFAIK is *not* deprecated, only its use as the build frontend is).

PR #836, opened 2024-07-13
Neuron support

emeryberger:


emeryberger:
Note that this PR changes - in fact, fixes - the way GPU utilization is reported. It is no longer normalized to the number of cores in use, but rather to the total number of GPU cores, which paints a better picture of how well code is using GPU (or here, Neuron) cores.

PR #838, opened 2024-07-16
Fix CPU-only profiling failing with "AttributeError: module 'pynvml' …

vavanade:
…has no attribute 'nvml'"

The error suppression does not work for an environment with `nvidia-ml-py==12.555.43` and `nvidia-ml-py3==7.352.0`


emeryberger:
Thanks for the PR! It was slightly incorrect but I've merged a corrected version.

PR #844, opened 2024-07-29
Improves magic handling

emeryberger:
Fixes https://github.com/plasma-umass/scalene/issues/843 by providing more robust magic detection and removal.

Issue #843, opened 2024-07-29
ast.parse crashes in certain edge cases


D-VR:
**Describe the bug**
It seems like in certain cases scalene will feed parts of comments into ast.parse, which then causes a crash.
This is more likely to happen if you increase the profiling frequency.

see two example crashes:

```
Scalene: An exception of type SyntaxError occurred. Arguments:
('unterminated triple-quoted string literal (detected at line 51)', ('<unknown>', 33, 9, '        """%s', 33, 9))
Traceback (most recent call last):
  File "user/venv/lib/python3.12/site-packages/scalene/scalene_profiler.py", line 2120, in run_profiler
    exit_status = profiler.profile_code(
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "user/venv/lib/python3.12/site-packages/scalene/scalene_profiler.py", line 1783, in profile_code
    did_output = Scalene.output_profile(left)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "user/venv/lib/python3.12/site-packages/scalene/scalene_profiler.py", line 838, in output_profile
    json_output = Scalene.__json.output_profiles(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "user/venv/lib/python3.12/site-packages/scalene/scalene_json.py", line 486, in output_profiles
    enclosing_regions = ScaleneAnalysis.find_regions(code_str)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "user/venv/lib/python3.12/site-packages/scalene/scalene_analysis.py", line 109, in find_regions
    tree = ast.parse(src)
           ^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/ast.py", line 52, in parse
    return compile(source, filename, mode, flags,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<unknown>", line 33
    """%s
    ^
SyntaxError: unterminated triple-quoted string literal (detected at line 51)
```

```
('unterminated string literal (detected at line 593)', ('<unknown>', 593, 17, "      Each group's index will be passed to the user defined function", 593, 17))
Traceback (most recent call last):
  File "user/venv/lib/python3.12/site-packages/scalene/scalene_profiler.py", line 2120, in run_profiler
    exit_status = profiler.profile_code(
                  ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #847, opened 2024-08-01
Replace magics with comments only when in Jupyter notebooks.

emeryberger:


emeryberger:
Replaces and closes https://github.com/plasma-umass/scalene/pull/844 and hopefully resolves https://github.com/plasma-umass/scalene/issues/843.

PR #859, opened 2024-08-25
Preliminary inclusion of time in tooltips

emeryberger:
Adds time to tooltips which appear when hovering over time % breakdowns.


emeryberger:
Fixes https://github.com/plasma-umass/scalene/issues/717.

PR #867, opened 2024-10-04
Fixed LD_LIBRARY_PATH squashing

sternj:
Because of how the current environment is checked, `LD_LIBRARY_PATH` and `LD_PRELOAD` are squashed if present, rather than prepended to. This commit prevents this.

Also preemptively fixed a potential issue where `PYTHONMALLOC` would not be properly squashed if present.

### DIFF
--- a/scalene/scalene_preload.py
+++ b/scalene/scalene_preload.py
@@ -34,28 +34,50 @@ class ScalenePreload:
 
         elif sys.platform == "linux":
             if args.memory:
-
-                sanitized_path = scalene.__path__[0]
+ 
+                unescaped_path = scalene.__path__[0]
+                escaped_path = unescaped_path.replace(" ", r"\ ")
                 if escape_spaces:
-                    # This function is used in two places, in `setup_preload` (where escaping spaces causes problems)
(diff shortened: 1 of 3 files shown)


sternj:
I think that what we'll need to do (rather soon) is escape all of the unescaped spaces in environment variables in `redirect_python`, if there are spaces in any environment variable that we don't mess with then they'll cause issues if the wrapper script is ever invoked.

emeryberger:
Any reason not to do this in the current PR?

sternj:
It's potentially a more far-reaching change, I can't think of any unintended effects but they may exist, especially given that it's happening at the invocation site for any subprocesses. What I have here is the minimal change to fix the issue at hand, but I'll see what the test suite says about the further-reaching solution.

sternj:
Note that this doesn't help with Windows environment variables, which I do not know the semantics of

emeryberger:
Scalene doesn't make specific use environment variables in Windows, should be fine, but should also be tested. Can we get Windows testing into the CI?

sternj:
I don't see why it isn't possible! I'll add it to the list.

PR #874, opened 2024-10-28
Running into an error where PyTorch is using

mmaloney007:
higher and higher SIGNAL codes…. I think this
could be fixed by updating the replace_signal
to take in additional codes


sternj:
@mmaloney007 thank you for the PR! Would you mind updating the title to make it a bit more clear and show in the description an example of the sort of undesirable behavior that the current implementation would give rise to that your implementation would fix?

emeryberger:
Bump @mmaloney007

sternj:
Testing this on my own, any signum that the `signal.Signals` enum marks as invalid is also invalid any signum that the `signal.Signals` enum marks as invalid is also invalid as an argument to `signal.signal`, at least on MacOS and Linux and at least for Python 3.10 to 3.13. If I missed something I'm glad to reopen, but given the staleness of this PR and the lack of any further information on what gave rise to it, I'm closing for now.

Issue #671, opened 2023-08-30
IndentationError: unexpected indent


fcappucci:
**Describe the bug**
I am trying to use scalene to improve a rather big program code, developed to re-grid satellite's data observations.
The code is running w/o error within the python environment (Python 3.11.4), the libraries ingested by the code are numpy, scipy, netCDF4, os, shutil. No parallel computation is implemented at current stage.

When I run the code via scalene I've got the following error message:

Scalene: An exception of type IndentationError occurred. Arguments:
('unexpected indent', ('<unknown>', 50, 4, '    Main class\n', 50, -1))
Traceback (most recent call last):
  File "/nethome/cappufa/miniconda3/envs/scalenv/lib/python3.11/site-packages/scalene/scalene_profiler.py", line 2118, in run_profiler
    exit_status = profiler.profile_code(
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/nethome/cappufa/miniconda3/envs/scalenv/lib/python3.11/site-packages/scalene/scalene_profiler.py", line 1904, in profile_code
    did_output = Scalene.output_profile()
                 ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nethome/cappufa/miniconda3/envs/scalenv/lib/python3.11/site-packages/scalene/scalene_profiler.py", line 945, in output_profile
    json_output = Scalene.__json.output_profiles(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nethome/cappufa/miniconda3/envs/scalenv/lib/python3.11/site-packages/scalene/scalene_json.py", line 415, in output_profiles
    enclosing_regions = ScaleneAnalysis.find_regions(code_str)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nethome/cappufa/miniconda3/envs/scalenv/lib/python3.11/site-packages/scalene/scalene_analysis.py", line 106, in find_regions
    tree = ast.parse(src)
           ^^^^^^^^^^^^^^
  File "/nethome/cappufa/miniconda3/envs/scalenv/lib/python3.11/ast.py", line 50, in parse
    return compile(source, filename, mode, flags,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<unknown>", line 50
    Main class
IndentationError: unexpected indent

**Desktop (please complete the following information):**
[...]

PR #875, opened 2024-10-30
Revised regex for magics removal

emeryberger:
The previous regex (meant to eliminate Jupyter notebook 'magics'), in some edge cases, would cause parsing to fail (leading to an `IndentationError`). This PR updates the regex to prevent that from happening.

### DIFF
--- a/scalene/scalene_analysis.py
+++ b/scalene/scalene_analysis.py
@@ -212,7 +212,7 @@ class ScaleneAnalysis:
             # The above line will fail if not running in a notebook,
             # in which case we return the original source unchanged.
             # Regular expression to match and replace magic commands with comments
-            source = re.sub(r'(^\s*)%{1,2}', r'\1#', source, flags=re.MULTILINE)
+            source = re.sub(r'(^\s*)%{1,2}(\w+)', r'\1# \2', source, flags=re.MULTILINE)
         finally:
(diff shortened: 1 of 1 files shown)


emeryberger:
Fixes https://github.com/plasma-umass/scalene/issues/671.

emeryberger:
For posterity: this code (from Python 3.12's `socket.py`) leads to the above-mentioned problem.

```
import _socket

class socket(_socket.socket):

    def __repr__(self):
        """Wrap __repr__() to reveal the real class name and socket                                                                                       
        address(es).                                                                                                                                      
        """
        closed = getattr(self, '_closed', False)
        s = "<%s.%s%s fd=%i, family=%s, type=%s, proto=%i" \
            % (self.__class__.__module__,
               self.__class__.__qualname__,
               " [closed]" if closed else "",
               self.fileno(),
               self.family,
               self.type,
               self.proto)
```

The culprit is the interaction between the backslash and the `%` on the following line. With the old regexp, part of this code was incorrectly transformed to the following (note that the `%` has been replaced by a `#`):

```
        s = "<%s.%s%s fd=%i, family=%s, type=%s, proto=%i" \
            # (self.__class__.__module__,
```

With the new regex, the `%` is correctly left unchanged.

PR #903, opened 2025-02-04
Fix module import issues for deeply nested modules (e.g., x.y.z).

glebarez:
Turns out the issue https://github.com/plasma-umass/scalene/issues/539 still persists for deeply nested modules, e.g.:
```
scalene -m x.y.z
```
This PR fixes this behaviour by enhancing the solution introduced in https://github.com/plasma-umass/scalene/pull/646


emeryberger:
Looks good - can you please provide a test case?

PR #920, opened 2025-05-09
Handle missing temp files during output without crashing

oklev:
Scalene currently crashes during output step after successful profiling if temp files have been removed; this makes it difficult to use with any libraries that clean their own temp files. Suggest try/except to still proceed even if a temp file is missing.

On second thought I'm not actually sure if this will fix the similar issue that I linked in the commit; but it did fix an issue I encountered where scalene would crash when running on a program that used pybedtools.

### DIFF
--- a/scalene/scalene_output.py
+++ b/scalene/scalene_output.py
@@ -373,7 +373,10 @@ class ScaleneOutput:
                     ),
                 )
 
-        null = tempfile.TemporaryFile(mode="w+")
+        try: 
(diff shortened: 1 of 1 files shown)


emeryberger:
Thanks for the PR. Is there a good reason to not just use `os.devnull` here all the time, instead of in response to an exception?

oklev:
Not that I know of! I'm not super familiar with the full function of this part of the codebase so I changed this the way I did mostly so that the behavior would unchanged unless it caused an exception.

Issue #805, opened 2024-04-07
Probable bug with scalene when profiling async code


dheerajck:
Python 3.11.6
Scalene version 1.5.38 (2024.03.24)

Profiled values arent in the correct place

To reproduce run this
`scalene --cpu-percent-threshold 0 async_await.py --cli`
and press Control + C after 3 seconds from when program prints "Task 1 done"
```
#  async_await.py
import asyncio


async def task():
    print('Task 1')
    await asyncio.sleep(5)
    a = 1
    a = a * 100
    print('Task 1 done')


async def gather():
    print('Gather 1')
    asyncio.create_task(task())
    print('Gather 1 done')


async def main():
    await asyncio.gather(gather())
    while True:
        await asyncio.sleep(1)

    print('Main done')


if __name__ == '__main__':
    try:
        asyncio.run(main())
    except KeyboardInterrupt:
        pass
```

output
```
/home/dheeraj/playground/profiler_test/async_await.py: % of time = 100.00% (5.009s) out of 5.009s.                                                       
       ╷       ╷       ╷       ╷        ╷       ╷               ╷       ╷                                                                                                                                       
       │Time   │–––––– │–––––– │Memory  │–––––– │–––––––––––    │Copy   │                                                                                                                                       
  Line │Python │native │system │Python  │peak   │timeline/%     │(MB/s) │/home/dheeraj/playground/profiler_test/async_await.py                                                                                  
╺━━━━━━┿━━━━━━━┿━━━━━━━┿━━━━━━━┿━━━━━━━━┿━━━━━━━┿━━━━━━━━━━━━━━━┿━━━━━━━┿━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╸
     1 │       │       │       │        │       │               │       │import asyncio                                                                                                                         
     2 │       │       │       │        │       │               │       │                                                                                                                                       
     3 │       │       │       │        │       │               │       │                                                                                                                                       
     4 │       │       │       │        │       │               │       │async def task():                                                                                                                      
     5 │       │       │       │        │       │               │       │    print('Task 1')                                                                                                                    
     6 │       │       │  [1]  │        │       │               │       │    await asyncio.sleep(5)                                                                                                             
[...]

PR #930, opened 2025-08-06
asyncio extension

bdunahu:
Currently, it is impossible for Scalene to report over 100% for a program; native code which executes in parallel should still be associated with a single line of python, and even if this were not the case (this applies to the multiprocessing library too), the samples can only be accounted for if they are present in the new_frames variable during each sample, each of which are currently treated as separate threads and assigned a normalized time based on how many there are.

This means the reporting problem is a new issue: it would be wrong to add the idle frames to new_frames, because that would imply they block the python interpreter from doing anything else while they're waiting.

When an asynchronous task is suspended, we instead assume it is waiting for the entire sampling interval. The solution this PR implements for the reporting problem is to treat idle tasks as if they run sequentially after non-waiting code, one after the other (i.e., nothing truly happens in parallel). This means the total CPU time passed is adjusted to match every sample.

Because asynchronous tasks run regardless of what the GIL is doing, the results are usually biased towards asynchronous code. It still leads to the behavior most users would likely expect.

Current state:
- idle-task frame collection logic is implemented in scalene_asyncio
- it is possible to prevent time from being assigned to the asyncio event loop by filtering out frames which belong to a thread that is running an event loop but has no current task. However, I do not do this, because throwing out frames complicates the above approach to reporting results.
- the passing of Scalene.should_trace is hacky. This function is also passed to scalene_utility.add_stack, is it possible for the implementation be moved to the utility file?

### DIFF
--- a/scalene/scalene_asyncio.py
+++ b/scalene/scalene_asyncio.py
@@ -24,16 +24,6 @@ class ScaleneAsyncio:
     current_task = None
 
     @staticmethod
-    def current_task_exists(tident) -> bool:
-        """Given TIDENT, returns true if a current task exists.  Returns
-        true if no event loop is running on TIDENT."""
-        current = True
-        loop = ScaleneAsyncio.loops.get(tident, None)
-        if isinstance(loop, asyncio.AbstractEventLoop):
-            current = asyncio.current_task(loop)
-        return bool(current)
-
-    @staticmethod
     def compute_suspended_frames_to_record(should_trace) -> List[FrameType]:
         """Collect all frames which belong to suspended tasks."""
         # TODO this is an ugly way to access the function
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -1006,15 +1006,6 @@ class Scalene:
         average_c_time = c_time / total_threads
         average_cpu_time = (python_time + c_time) / total_threads
 
-        # filter out all frames which are not running a current task.
-        # this is done after calculating the total frames (threads), because
-        # an idle thread still takes up CPU time.
-        new_frames = [
-            (frame, tident, orig_frame)
-            for frame, tident, orig_frame in new_frames
-            if ScaleneAsyncio.current_task_exists(tident)
-        ]
-
         # Now attribute execution time.
         # First, handle the main thread.
         if (new_frames):


scalene/scalene_profiler.py
```diff
@@ -1042,7 +1045,7 @@ def process_cpu_sample(
             Scalene.__stats.gpu_stats.gpu_mem_samples[fname][lineno].push(gpu_mem_used)
 
         # Now handle the rest of the threads.
-        for frame, tident, orig_frame in new_frames:
+        for frame, tident, orig_frame in new_frames + async_frames:
```
jaltmayerpizzorno, line 1048: You only use `async_frames` here, yet `total_frames` and the averages used to attribute time are computed across `new_frames`... is that right/consistent?

scalene/scalene_asyncio.py
```diff
@@ -0,0 +1,205 @@
...
+                frame = frame.f_back
+        return None
+
+    @staticmethod
+    def _get_frames_from_loops(loops) -> \
+            List[Tuple[FrameType, int, FrameType]]:
+        """Given LOOPS, returns a flat list of frames corresponding to idle
+        tasks."""
+        return [
+            (frame, tident, None) for loop, tident in loops
```
jaltmayerpizzorno, line 86: So each task frame gets returned with the `tident` of the thread running it?  Did you review how `tident` is used in the code to see if that still makes sense for task frames?

scalene/scalene_profiler.py
```diff
@@ -1068,10 +1071,12 @@ def process_cpu_sample(
                 # Ignore sleeping threads.
```
jaltmayerpizzorno, line 1071: Will this work correctly for task frames showing the `tident` of the loop processing thread?

jaltmayerpizzorno (commented):
Preliminary review while I wait for an answer to a question I posed on Slack.

emeryberger:
With the recent refactoring of scalene_profiler.py, this is going to take some work to bring up to date.

bdunahu:
> With the recent refactoring of scalene_profiler.py, this is going to take some work to bring up to date.

Is there room for improving how Scalene reports asyncio code? Attributing CPU time to the event loop internals is not very useful to the user, but it is where the CPU actually spends time when the event loop sits in the select call with nothing to do (https://github.com/plasma-umass/scalene/issues/805 shows how the correct results are still un-intuitive).

There didn't seem to be an easy way to work in 'asynchronous time' into the profile results like other profilers do unless we are considering adding a new column/flag? If not, this pull request can probably be closed.

emeryberger:
Thanks for this contribution! Since this PR was opened, Scalene has landed async profiling support on master that takes a different (and more comprehensive) architectural approach:

- **`scalene_async.py`** with two instrumentation strategies: polling via `asyncio.all_tasks()` on Python 3.9+, and precise `sys.monitoring` PY_YIELD/PY_RESUME tracking on Python 3.12+
- **Dedicated `AsyncStatistics`** class that tracks await time, task names, and concurrency per-line separately (rather than attributing to C-time)
- **Separate signal queue** for async samples, keeping the async profiling path cleanly decoupled from CPU sample processing
- Full integration into the output, JSON, and statistics pipeline with an `--async` CLI flag (enabled by default)

Given the overlap, we're closing this PR. We did pick up the `#!/usr/bin/env bash` portability fix from your `redirect_python.py` change — thanks for that!

emeryberger:
Also, would be great if you could take a look at the current async support; if you find any issues or have suggestions, let me know! (Not mentioned above, Scalene now tracks "wait time" in the UI for profiling async code.)

bdunahu:
Hi @emeryberger,

That all sounds wonderful! I'm sorry for going quiet, I have been overwhelmed by my employment search since February and had stopped thinking about this.

I will put it on my todo list and try to check out the new features and see if I can break them and come up with any nitpicks relatively soon, since a lot of that sounds familiar to me.

PR #931, opened 2025-08-08
Enable output to both JSON and CLI

zyrxvo:
**Context:**
The CLI output, especially with the `--reduced-profile` flag, provides concise, actionable, and human readable information for improving and optimizing code. Additionally, the JSON output provides verbose details, perfect to use for machine-automated interpretation and analysis of the results.

**Changes:**
This PR enables the ability to save _both_ the CLI output and the JSON output from the same profile when both the `--cli` and `--json` flags are used in conjunction with `--outfile`.

**Drawbacks:**
In order to prevent overwriting the JSON file, the plain text file is forced to have a `.txt` suffix. For consistency the HTML files are all forced to have a `.html` suffix. 

**Comments:**
I understand that the default behaviour is to produce a JSON file _and_ an HTML file for viewing in the browser. The change in this PR reflects the desire to provide the option of having the human readable profile as a text file, rather than only with a browser.

### DIFF
--- a/.github/workflows/build-and-upload.yml
+++ b/.github/workflows/build-and-upload.yml
@@ -30,14 +30,14 @@ jobs:
     strategy:
       matrix:
         python_version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
-        plat: ['manylinux_2_28', 'macos-13', 'macos-latest', 'windows-latest']
+        plat: ['manylinux_2_28', 'macos-latest', 'windows-latest']
         include:
           - plat: manylinux_2_28
             os: ubuntu-latest
             container: quay.io/pypa/manylinux_2_28_x86_64  # https://github.com/pypa/manylinux
 
-          - plat: macos-13
-            os: macos-13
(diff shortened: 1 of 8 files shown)


emeryberger:
Please address the CI issues above, thanks!

zyrxvo:
I addressed the CI issues by trying to keep the behaviour as similar to the original as possible. The only difference should be with `--cli` in combination with `--outfile`. Now, when `--json` is included, it also saves the table summary, either as it would be with only `--cli` or with `--cli` and `--html`.

zyrxvo:
I'm open to discussion on the merit of these suggestions and their alignment with your vision of `scalene`.

emeryberger:
This is now obsoleted by the new CLI. It always generates a JSON file (`scalene run`) and then you can generate a CLI view afterwards (`scalene view --cli`).

zyrxvo:
Thank you so much! The new CLI is great. It's very flexible and very powerful. I look forward to using it on my projects.

PR #940, opened 2025-08-20
Cpu Timestamp + Visualization Changes

nehamur:
- Records CPU time samples every time scalene is run
- Records absolute starting time and the start time of the first perf


emeryberger:
Merging in the interest of expedience but I don't think the merging code should be needed.

PR #946, opened 2025-08-31
Fix signal timer resetting

sternj:
Fixed user timer being disabled after it "rings" instead of being reset

### DIFF
--- a/.github/workflows/test-smoketests.yml
+++ b/.github/workflows/test-smoketests.yml
@@ -51,6 +51,12 @@ jobs:
     - name: decorator smoke test
       run: python test/smoketest_profile_decorator.py
 
+      # NOTE: this is a regression test for signals not being
+      # restarted properly. It checks whether the program 
+      # halts or runs forever, not for correctness of timing.
+    - name: signal smoketest
+      run: python test/smoketest.py test/signal_test.py
+      timeout-minutes: 0.5
     # FIXME: these tests are broken under the current Github runner
     #
     # - name: line invalidation test
--- a/scalene/scalene_profiler.py
+++ b/scalene/scalene_profiler.py
@@ -731,7 +731,6 @@ class Scalene:
                         to_wait = min(remaining_time, next_interval)
                     else:
                         to_wait = next_interval
-                        Scalene.client_timer.reset()
                     Scalene.__signal_manager.restart_timer(to_wait)
                 else:
                     Scalene.__signal_manager.restart_timer(next_interval)
--- a/test/signal_test.py
+++ b/test/signal_test.py
@@ -1,0 +1,16 @@
+import signal
+
+iterations = 10
+
+def my_handler(sig, frame):
+    global iterations
+    print(f"seconds remaining: {iterations}")
+    if iterations > 0:
+        iterations -= 1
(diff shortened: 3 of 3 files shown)


.github/workflows/test-smoketests.yml
```diff
@@ -51,6 +51,12 @@ jobs:
     - name: decorator smoke test
       run: python test/smoketest_profile_decorator.py
 
+      # NOTE: this is a regression test for signals not being
+      # restarted properly. It checks whether the program 
+      # halts or runs forever, not for correctness of timing.
+    - name: signal smoketest
+      run: python test/smoketest.py test/signal_test.py
```
emeryberger, line 58: This is a bad failure case, plus we have a regression test suite in `tests/`. Make it a pytest and put it in that directory so it gets run in that workflow, but with a timeout for failure.

scalene/scalene_profiler.py
```diff
@@ -731,7 +731,6 @@ def cpu_signal_handler(
                         to_wait = min(remaining_time, next_interval)
                     else:
                         to_wait = next_interval
-                        Scalene.client_timer.reset()
```
emeryberger, line 734: This works but I still get the warning about the signal, so we need to filter that.

```
WARNING: Scalene uses Alarm clock: 14 to profile.
If your code raises Alarm clock: 14 from non-Python code, use SIGUSR1.
Code that raises signals from within Python code will be rerouted.
```

emeryberger (changes requested):

.github/workflows/test-smoketests.yml
sternj: I think the failure case is the right one-- the behaviors this patch was in response to were the signal not being delivered for whatever reason and this test checks for progress. 

I'll put it in the `tests` workflow, I actually think it makes sense as a `task` with a timeout rather than as a pytest though, unless you wanted to start making pytests that spawn new processes (which it doesn't seem like we do at the moment, the pytest suite seems to be mostly unit tests). I do think it's structurally more similar to the smoketests, since we're just checking "are the signals delivered and does it create a profile with any data in it", but that's your call

emeryberger: I mean that the CI will get hung up - no idea what the timeout is for that thing - so the latter suggestion works.

sternj: The timeout of the CI is 15 minutes and I put in a timeout of 30 seconds for the task I created, where would the hangup happen?

scalene/scalene_profiler.py
sternj: Right-- we print the warning because, while we can transparently redirect everything in Python-land, we want to warn the user in case they're doing something in C-land

.github/workflows/test-smoketests.yml
sternj: Also [the smoketests worked fine](https://github.com/plasma-umass/scalene/actions/runs/17362592575/job/49284657107?pr=946), I could make sure the timeout works as expected but I trust the `timeout-minutes` directive to do the right thing

emeryberger:
I'm going to merge; we can address the testing question later.

PR #1058, opened 2026-06-05
Fix #1056: signal-free CPU sampling on Apple Silicon with SME (M4+)

### LINKED ISSUE #1056: Numerical corruption on Mac M4+ chips

emeryberger:
## Problem

On Apple **M4+** chips (which implement ARM **SME**), running Scalene against numpy/Accelerate BLAS code silently **corrupts numerical results** — a by-construction-Hermitian matrix comes out non-Hermitian (max deviation jumps from ~1e-12 to ~1e3). Reported in #1056.

A diagnostic probe matrix run on **real M4 hardware** (macOS 26.5, FEAT_SME=1) pinned the root cause:

| probe | result |
|---|---|
| baseline (no signals) | clean |
| `SIG_IGN` (no handler invoked) | **clean** |
| empty **C** handler | **CORRUPT** |
| C handler on alternate stack | **CORRUPT** |
| Python no-op handler | **CORRUPT** |
| SIGPROF / SIGALRM variants | **CORRUPT** |

Only `SIG_IGN` is clean; an *empty* C handler corrupts as hard as a Python one, on any stack, for every signal/timer type. **The macOS kernel mis-saves/restores SME streaming state on `sigreturn`** — so no handler-side change can fix it. (Distinct from numpy #28687, which was spurious FP *warnings* fixed in numpy 2.3; this is real data corruption, still present with numpy 2.3.2.)

## Fix

Stop delivering profiling signals on these machines. Gated on `sysctl hw.optional.arm.FEAT_SME == 1`:

- **`scalene_mac_sampler.py`** (new): `MacThreadSampler` drives the CPU handler from a background thread on a wall-clock cadence (mirrors the existing Windows `windows_timer_loop`); no timer signal armed.
- **`scalene_utility.py`**: `is_apple_silicon_sme()` cached detection.
- **`scalene_signal_manager.py`**: `enable_signals(use_thread_sampler=...)` skips `setitimer`; `restart_timer` no-ops; `stop_mac_sampler()`. **Memory profiling preserved** by SIG_IGN'ing malloc/free/memcpy and poll-draining the sample queues from the sampler thread — safe because the native allocator writes each sample to its file *before* `raise()`.
- **`scalene_cpu_profiler.py`**: `thread_sampled` flag makes Python-vs-C attribution use the CALL\*-bytecode classifier (already used for worker threads) instead of virtual-timer deferral.
- **`scalene_profiler.py`**: detects SME, routes to the sampler, disables `--stacks` (with a warning) since its SIGPROF-to-all-threads would re-introduce the corruption.

**Non-SME platforms are completely unaffected** — the timer-signal path is unchanged.

## Testing

- New `tests/test_mac_sampler.py` (7 tests): sampler cadence, exception resilience, idempotent stop, memory poll-draining, SME detection.
- Full suite: 372 passed (1 pre-existing unrelated failure).
- On M2 (no SME): normal path unchanged; forced-SME path verified end-to-end — CPU + memory sampling work, and a mixed workload attributes matmul lines to C, Python loops to Python.

## ⚠️ Still needs M4 sign-off

The final *corruption-is-gone* proof requires SME hardware (maintainer's machine is M2). A one-command verifier (`verify_fix.sh`) and the diagnostic kit are prepared for that run. Opening now so **CI exercises the non-SME paths** across the support matrix while M4 verification is in flight.

Closes #1056 (pending M4 verification).

### DIFF
--- a/scalene/scalene_cpu_profiler.py
+++ b/scalene/scalene_cpu_profiler.py
@@ -32,7 +32,11 @@ class ScaleneCPUProfiler:
     """Handles CPU profiling sample processing."""
 
     def __init__(
-        self, stats: ScaleneStatistics, available_cpus: int, use_virtual_time: bool
+        self,
+        stats: ScaleneStatistics,
+        available_cpus: int,
+        use_virtual_time: bool,
+        thread_sampled: bool = False,
     ) -> None:
         """Initialize the CPU profiler.
 
@@ -41,10 +45,18 @@ class ScaleneCPUProfiler:
             available_cpus: Number of available CPUs for utilization calculations.
             use_virtual_time: Whether we're using virtual time (SIGVTALRM) or
                 wall clock time (SIGALRM) for CPU sampling.
+            thread_sampled: True when CPU samples are produced by a background
+                helper thread on a wall-clock cadence rather than a timer
+                signal (Apple SME path, issue #1056). In that mode there is no
+                virtual-timer deferral to exploit, so Python-vs-C time is
+                attributed purely from the bytecode at f_lasti (a CALL* opcode
+                means we're in a native call) — exactly as wall-clock mode
+                already does, regardless of the --use-virtual-time flag.
         """
         self._stats = stats
         self._available_cpus = available_cpus
         self._use_virtual_time = use_virtual_time
+        self._thread_sampled = thread_sampled
 
     def process_cpu_sample(
         self,
@@ -230,15 +242,24 @@ class ScaleneCPUProfiler:
                 # call, and ALL time (including python_time) should go to native.
                 # This handles the case where elapsed.virtual ≈ last_cpu_interval
                 # (c_time ≈ 0) but we're still returning from C code.
+                #
+                # The thread-sampled (Apple SME) path always uses this
+                # bytecode-based test: it has no virtual-timer deferral to
+                # measure c_time from, so the CALL* opcode at f_lasti is the
+                # sole signal for "this sample landed in a native call". This
+                # is the same classifier already used for worker threads.
                 is_at_call = ScaleneFuncUtils.is_call_function(
                     main_thread_frame.f_code,
                     ByteCodeIndex(main_thread_frame.f_lasti),
                 )
 
-                if not self._use_virtual_time and is_at_call:
-                    # Wall clock mode at a CALL instruction: attribute all time
-                    # to native on this line. The signal was deferred during the
-                    # C call, so even python_time was spent in C, not Python.
+                use_call_attribution = (
+                    not self._use_virtual_time or self._thread_sampled
+                )
+                if use_call_attribution and is_at_call:
+                    # At a CALL instruction: attribute all time to native on
+                    # this line. The sample landed inside the C call, so even
+                    # python_time was spent in C, not Python.
                     self._update_main_thread_stats(
                         fname,
                         lineno,
--- a/scalene/scalene_mac_sampler.py
+++ b/scalene/scalene_mac_sampler.py
@@ -1,0 +1,141 @@
+"""Signal-free CPU sampler for Apple Silicon with SME (issue #1056).
+
+On Apple M4+ chips the macOS kernel mis-saves/restores ARM SME
+("streaming"-mode) register state whenever a signal handler is invoked
+while a thread is mid-kernel in Accelerate BLAS. This silently corrupts
+in-flight numerical results — even an empty C signal handler triggers it
+(see scalene/issue-1056-probes and issue #1056). No handler-side change
+can fix it, because the corruption is applied by the kernel on
+``sigreturn``, after the handler returns.
+
+The only correctness-preserving mitigation is to stop delivering the
+CPU-profiling timer signal on these machines. This module replaces the
+``setitimer``/``SIGVTALRM`` mechanism with a background *helper thread*
+that periodically calls Scalene's CPU sample handler directly. Because no
+signal is ever delivered, the kernel never runs its buggy SME save/
+restore path, and the worker thread's matrix state stays intact.
+
+This mirrors the existing Windows approach (``windows_timer_loop`` in
+``scalene_signal_manager``), which already drives the CPU handler from a
+background thread for a different reason (``signal.raise_signal`` is
+unavailable off the main thread on Windows). The handler reads thread
+state via ``sys._current_frames()``, which is thread-safe and works from
+any thread, so calling it from this helper is sound.
+
+Python-vs-C time attribution still works: the CPU profiler classifies the
+main thread's time from the bytecode at ``f_lasti`` (a ``CALL*`` opcode
+means the frame is sitting in a native call) — the same rule already used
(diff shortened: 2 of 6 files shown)


emeryberger:
## CI note: the one red check (`ubuntu-latest, 3.13t`) is a pre-existing timeout, not a test failure

**38/39 checks pass**, including every macOS config — the only place this fix's code path can even activate.

The `run-tests (ubuntu-latest, 3.13t)` job fails on the **45-minute job timeout**, not an assertion:

> The job has exceeded the maximum execution time of 45m0s … The operation was canceled.

Reran it once; it timed out again at exactly 45m15s. Evidence this is unrelated to this PR:

- **The new code path is inert in CI.** It's gated on `is_apple_silicon_sme()` (`sysctl hw.optional.arm.FEAT_SME == 1`), which is `False` on all GitHub runners (Linux, and M1 macOS). So `use_thread_sampler`/`thread_sampled` are never `True` in CI — the timer-signal path runs exactly as before. The 7 new tests add ~1.5s total.
- **It's a master-wide slowdown, not this branch.** The last green master `tests.yml` run (May 12) finished `ubuntu 3.13t` in **~23 min**; the suite has since grown to where this slowest free-threaded config exceeds the 45m ceiling. The most recent master run failed on a *different* job (`3.14`) with the rest fail-fast-cancelled.

So this is the test suite outgrowing its CI timeout budget on the slowest free-threaded config — a separate infra issue (raise `timeout-minutes` in `tests.yml`), not a regression from this change.

⚠️ Reminder: the actual *corruption-is-gone* verification still requires real M4 (SME) hardware — CI only exercises the unaffected non-SME paths. A `verify_fix.sh` harness is prepared for that run.

timmysilv:
I don't see `verify_fix.sh`, but for what it's worth, I just validated the fix with `numerical_bug.py` from my initial bug report on my M5 Mac (trimmed terminal interactions to highlight key results):

```bash
# first, install scalene from PyPI and reproduce failure
$ uv pip install scalene
$ scalene run --cpu-only --no-stacks numerical_bug.py 300 1536 1536
iters=300 dim=1536 rank=1536 elapsed=51.24s max_hermitian_deviation=1.501e+02 (worst@iter 112) violations(>1e-06)=446 nan_or_inf=False
*** NUMERICAL CORRUPTION DETECTED: a by-construction-Hermitian matrix came out non-Hermitian ***

# then, install from fix commit and ensure success
$ uv pip install git+https://github.com/plasma-umass/scalene@7a095bec5e2e16613e16932cda1d7b3f94b66498
$ scalene run --cpu-only --no-stacks numerical_bug.py 300 1536 1536
iters=300 dim=1536 rank=1536 elapsed=51.44s max_hermitian_deviation=1.506e-12 (worst@iter 274) violations(>1e-06)=0 nan_or_inf=False
clean: results Hermitian to within tolerance (no corruption observed in this run)
```

I even ran without `--cpu-only` and it worked fine. This is great, thank you!

PR #1087, opened 2026-07-31
Fix UnicodeEncodeError for non-ASCII file paths (#1086)

### LINKED ISSUE #1086: UnicodeEncodeError when non-ASCII characters are in any imported file path
**Describe the bug**

When running scalene on a file that has a module in its import tree which is at a path with non-ascii characters in it, it will fail with a `UnicodeEncodeError`. The actual location is a little flaky, but most of the time the stacktrace looks like this (shortened for readability, full log with repro below):

```
$ uv run scalene run src/profiling_example.p
y 
 'ascii' codec can't encode character '\xfc' in position 79: ordinal not in range(128)
  File "/projects/[...]/.venv/lib/python3.12/site-packages/scalene/scalene_profiler.py", line 1523, in profile_code
[...]
  File "/projects/[...]/.venv/lib/python3.12/site-packages/scalene/scalene_tracing.py", line 141, in _passes_exclusion_rules
    the_path = sysconfig.get_path(p, n)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/sysconfig.py", line 672, in get_path
    return get_paths(scheme, vars, expand)[name]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/sysconfig.py", line 662, in get_paths
    return _expand_vars(scheme, vars)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/sysconfig.py", line 304, in _expand_vars
    res[key] = os.path.normpath(_subst_vars(value, vars))
                                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/sysconfig.py", line 275, in _subst_vars
    return s.format(**local_vars)
           ^^^^^^^^^^^^^^^^^^^^^^
UnicodeEncodeError: 'ascii' codec can't encode character '\xfc' in position 79: ordinal not in range(128)

```

This does NOT happen when passing `--cpu-only`. In that case, profiling works fine.

Scalene does still write a `scalene_profile.json` despite this, but it can't be displayed, the resulting HTML is nearly empty.

The error can sometimes also surface in an unrelated location, for example:

```
  File "/projects/[...]/optionale_verlängerung.py", line 202, in <module>
    class OptVerlOutputVt(FrozenBaseModel):
  File "/projects/[...]/.venv/lib/python3.12/site-packages/pydantic/_internal/_model_construction.py", line 255, in __new__
    complete_model_class(
  File "/projects/[...]/.venv/lib/python3.12/site-packages/pydantic/_internal/_model_construction.py", line 680, in complete_model_class
    cls.__pydantic_serializer__ = SchemaSerializer(schema, core_config)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.SchemaError: Error building `definitions` serializer:
  SchemaError: Error building `model` serializer:
  SchemaError: Error building `model-fields` serializer:
  UnicodeEncodeError: 'ascii' codec can't encode character '\xe4' in position 92: ordinal not in range(128)
```

**To Reproduce**
1. Make sure `uv` is installed
2. Download the repro-script from https://gist.github.com/iFreilicht/5c1710cc3f862c47d818c9cc03a29c3c
3. Run `bash scalene_repro.sh`
4. See following output

<details>
<summary>Output of `bash scalene_repro.sh`</summary>

```
$ bash scalene_repro.sh
### BROKEN: memory tracker active — expect UnicodeEncodeError / SystemError / SIGSEGV
 'ascii' codec can't encode character '\xfc' in position 60: ordinal not in range(128)
  File "/home/gitlab-workspaces/.cache/uv/archive-v0/QBonbADp3FUl6J_o/lib/python3.12/site-packages/scalene/scalene_profiler.py", line 1559, in profile_code
  File "/home/gitlab-workspaces/.cache/uv/archive-v0/QBonbADp3FUl6J_o/lib/python3.12/site-packages/scalene/replacement_exec.py", line 122, in exec_replacement
    orig_exec(__source, globals, locals, **kwargs)
  File "/projects/[...]/work/main.py", line 2, in <module>
    print(überschüsse.f())
          ^^^^^^^^^^^^^^^
  File "/projects/[...]/work/überschüsse.py", line 4, in f
    s += i
    ^
  File "/home/gitlab-workspaces/.cache/uv/archive-v0/QBonbADp3FUl6J_o/lib/python3.12/site-packages/scalene/scalene_profiler.py", line 960, in cpu_signal_handler
    frames = compute_frames_to_record(Scalene._should_trace)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gitlab-workspaces/.cache/uv/archive-v0/QBonbADp3FUl6J_o/lib/python3.12/site-packages/scalene/scalene_utility.py", line 268, in compute_frames_to_record
    while not should_trace(Filename(fname), func):
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gitlab-workspaces/.cache/uv/archive-v0/QBonbADp3FUl6J_o/lib/python3.12/site-packages/scalene/scalene_profiler.py", line 1256, in _should_trace
    return Scalene.__tracing.should_trace(filename, func)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gitlab-workspaces/.cache/uv/archive-v0/QBonbADp3FUl6J_o/lib/python3.12/site-packages/scalene/scalene_tracing.py", line 100, in should_trace
    if not self._passes_exclusion_rules(filename):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gitlab-workspaces/.cache/uv/archive-v0/QBonbADp3FUl6J_o/lib/python3.12/site-packages/scalene/scalene_tracing.py", line 141, in _passes_exclusion_rules
    the_path = sysconfig.get_path(p, n)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/sysconfig.py", line 672, in get_path
    return get_paths(scheme, vars, expand)[name]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/sysconfig.py", line 662, in get_paths
    return _expand_vars(scheme, vars)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/sysconfig.py", line 304, in _expand_vars
    res[key] = os.path.normpath(_subst_vars(value, vars))
                                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/sysconfig.py", line 275, in _subst_vars
    return s.format(**local_vars)
           ^^^^^^^^^^^^^^^^^^^^^^
UnicodeEncodeError: 'ascii' codec can't encode character '\xfc' in position 60: ordinal not in range(128)
Scalene: The specified code did not run for long enough to profile.
By default, Scalene only profiles code in the file executed and its subdirectories.
To track the time spent in all files, use the `--profile-all` option.
exit code: 1

### WORKS: --cpu-only (memory tracker / pywhere disabled)
4499998500000

Scalene: profile saved to /projects/[...]/work/out_cpu.json
  To view in browser:  scalene view /projects/[...]/work/out_cpu.json
  To view in terminal: scalene view --cli /projects/[...]/work/out_cpu.json
exit code: 0

### WORKS: cProfile on the same code (handles Unicode paths natively)
4499998500000
exit code: 0

Note: the broken run is flaky. Under light load it may occasionally pass;
re-run a few times, or import more umlaut-named modules, to see it fail reliably.
```

</details>

**Expected behavior**

I would expect scalene to properly handle filepaths with non-ascii characters depending on the filesystem encoding

**Potential cause**

I believe the issue is in one of these locations: https://github.com/plasma-umass/scalene/blob/c2e9e17b2ae499cdcdf549a0f12862d8307634f5/src/source/pywhere.cpp#L226-L227, https://github.com/plasma-umass/scalene/blob/c2e9e17b2ae499cdcdf549a0f12862d8307634f5/src/source/pywhere.cpp#L463-L464 and https://github.com/plasma-umass/scalene/blob/c2e9e17b2ae499cdcdf549a0f12862d8307634f5/src/source/pywhere.cpp#L942-L946 which all call `PyUnicode_AsASCIIString` on filepaths.

This should probably be `PyUnicode_EncodeFSDefault` or `PyUnicode_AsUTF8`, I'm not 100% sure. Even then it's kinda risky to let these run without a null-check because from what I understand, these functions set an exception in the python interpreter and just return `NULL` on failure, which can cause the error to surface in an unrelated location.

 - OS: Ubuntu Linux
 - Browser: N/A
 - Version: 2.3.0, also reproduced on current `master` (https://github.com/plasma-umass/scalene/commit/c2e9e17b2ae499cdcdf549a0f12862d8307634f5)

emeryberger:
Fixes #1086.

## Root cause

`pywhere.cpp` converted `code->co_filename` with **`PyUnicode_AsASCIIString`** on three paths: the stack walker used by the allocator interposer, `on_stack()`, and the settrace line callback. For any path containing a non-ASCII character — e.g. a module at `.../überschüsse.py` — that call fails, returning `NULL` *and leaving a `UnicodeEncodeError` set on the thread*.

Since these run from the native allocator hook and from trace callbacks, the stray exception surfaced later at an arbitrary, unrelated Python line — which is why the reporter saw it land in `sysconfig.get_path` on one run and inside pydantic on the next, and why the failure looked flaky. Two of the sites also fed the `NULL` straight into `PyBytes_AsString`/`strstr`, giving the reported SIGSEGV.

The Python side compounded it: `ScaleneMapFile.get_str()` decoded the native sample records as ASCII, so any allocation attributed to such a file also blew up the sample-draining loop.

Only `--memory` runs were affected, matching the reporter's observation that `--cpu-only` worked — that's the mode that loads the native interposer.

## Changes

- **`src/source/pywhere.cpp`** — new `encodeFilename()` helper encoding as UTF-8 with `surrogateescape`. That's the exact inverse of how CPython decodes filesystem paths, so every path Python can represent — including undecodable bytes, which appear as lone surrogates — round-trips. It clears any exception on failure and returns `nullptr`; all three call sites now NULL-check. Also removes a dead encode-and-discard in `trace_func`.
- **`src/include/traceconfig.hpp`** — the package-path encode used `"strict"` and leaked a pending exception on failure; now `surrogateescape` + `PyErr_Clear()`.
- **`scalene/scalene_mapfile.py`** — new `decode_sample()` using the matching UTF-8/`surrogateescape` pair, replacing the ASCII decodes.

## Test

`tests/test_issue1086_non_ascii_paths.py` is picked up by the existing `python3 -m pytest` step in `tests.yml`, so it runs on Ubuntu + macOS across Python 3.9–3.14. Three cases:

1. `decode_sample` directly — deterministic, including lone surrogates.
2. + 3. End-to-end `scalene run --memory` on a program importing a non-ASCII module, with the non-ASCII component in the filename and in a parent directory. Asserts no Unicode error, the program ran to completion, exit code 0, and the path comes back through the profile intact (NFC-normalized comparison so an NFD-normalizing filesystem doesn't false-fail).

The crash assertions are hard on every attempt; only the "file appears in the profile" check retries against the usual sampling flake.

## Verification

- Reproduced the issue on macOS / Python 3.14: 3/3 runs died with `'ascii' codec can't encode character '\xfc'`, same traceback shape as the report.
- With the fix, the profile is complete (`überschüsse.py`: 20 MB malloc, 89% CPU); `--memory --stacks` also carries the non-ASCII path correctly through the native stack buffer.
- Reverted **only** the native fix and rebuilt → the two end-to-end tests fail as intended; restored → pass.
- Full `tests/` suite: 445 passed, 13 skipped. `ruff` clean; `mypy` shows only the pre-existing `scalene_signal_manager.py` error also present on master.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scalene/scalene_mapfile.py
+++ b/scalene/scalene_mapfile.py
@@ -9,6 +9,22 @@ if sys.platform != "win32":
 Filename = NewType("Filename", str)
 
 
+def decode_sample(raw: bytes) -> str:
+    """Decode one sample record written by the native side.
+
+    Records embed source filenames, which the native side encodes as UTF-8
+    with surrogate escapes (see ``encodeFilename`` in
+    ``src/source/pywhere.cpp``). Decoding the same way is the exact inverse,
+    so paths with non-ASCII characters -- and even paths with undecodable
+    bytes, which CPython represents as lone surrogates -- round-trip.
+
+    This used to decode as ASCII, which raised ``UnicodeDecodeError`` and
+    took down the sample-processing loop for any allocation attributed to a
+    file whose path contained e.g. an umlaut (issue #1086).
+    """
+    return raw.decode("utf-8", errors="surrogateescape")
+
+
 class ScaleneMapFile:
 
     # Things that need to be in sync with the C++ side
@@ -249,7 +265,7 @@ class ScaleneMapFile:
             self._buf[length:] = b"\x00" * (self.MAX_BUFSIZE - length)
 
             # Validate the sample has expected format before accepting
-            sample_preview = self._buf[:100].decode("ascii", errors="replace")
+            sample_preview = self._buf[:100].decode("utf-8", errors="replace")
             if "," not in sample_preview:
                 # Malformed sample - skip it but advance position
                 self._lastpos = struct.pack("<Q", end_pos)  # type: ignore[assignment]
@@ -264,4 +280,4 @@ class ScaleneMapFile:
 
     def get_str(self) -> str:
         """Get the string from the buffer."""
-        return self._buf.rstrip(b"\x00").split(b"\n")[0].decode("ascii")
+        return decode_sample(bytes(self._buf.rstrip(b"\x00").split(b"\n")[0]))
--- a/src/include/traceconfig.hpp
+++ b/src/include/traceconfig.hpp
@@ -74,11 +74,17 @@ class TraceConfig {
     if (scalene_pkg_path_obj && scalene_pkg_path_obj != Py_None) {
       scalene_pkg_path_owner = scalene_pkg_path_obj;
       Py_IncRef(scalene_pkg_path_owner);
-      auto enc =
-          PyUnicode_AsEncodedString(scalene_pkg_path_obj, "utf-8", "strict");
+      // surrogateescape (not strict): a path holding undecodable bytes shows
+      // up in Python as lone surrogates, which a strict encoder rejects —
+      // leaving an exception pending that would later surface at an
+      // unrelated line (issue #1086).
+      auto enc = PyUnicode_AsEncodedString(scalene_pkg_path_obj, "utf-8",
+                                           "surrogateescape");
       if (enc) {
         scalene_pkg_path = std::string(PyBytes_AsString(enc));
         Py_DecRef(enc);
+      } else {
+        PyErr_Clear();
       }
     } else {
       scalene_pkg_path_owner = nullptr;
--- a/src/source/pywhere.cpp
+++ b/src/source/pywhere.cpp
@@ -172,6 +172,37 @@ PyObject* set_scalene_done_false(PyObject* self, PyObject* args) {
   Py_RETURN_NONE;
 }
 
+// Encode a Python filename (a ``str``, e.g. ``code->co_filename``) as
+// bytes we can hand to C string functions.
+//
+// This used to be ``PyUnicode_AsASCIIString``, which fails for any path
+// containing a non-ASCII character — a very ordinary situation (issue
+// #1086: a module at ``.../überschüsse.py``). Two things went wrong there:
+// the caller lost the filename, and — worse — the failed conversion left a
+// ``UnicodeEncodeError`` set on the thread. Since this code runs from the
+// allocator interposer and from trace callbacks, that stray exception
+// surfaced later at an arbitrary, unrelated Python line (or as a
+// ``SystemError``/crash), which is why the bug looked flaky.
+//
+// UTF-8 with ``surrogateescape`` is the right encoding here: it is the
+// inverse of how CPython decodes filesystem paths, so every path that
+// Python can represent — including undecodable bytes, which appear as lone
+// surrogates — round-trips. The Python side decodes these records the same
+// way (see ``decode_sample`` in scalene/scalene_mapfile.py). On failure we
+// return nullptr with no exception left pending; callers must NULL-check.
+static PyObject* encodeFilename(PyObject* unicode) {
+  if (unicode == nullptr) {
+    return nullptr;
+  }
+  PyObject* bytes =
+      PyUnicode_AsEncodedString(unicode, "utf-8", "surrogateescape");
+  if (bytes == nullptr) {
+    // Non-str object, or an encoder failure we can't do anything about.
+    PyErr_Clear();
+  }
+  return bytes;
+}
+
 // Core stack-walk used by both whereInPython and whereInPythonWithStack.
 // If ``stack_buf`` is non-null and ``stack_buf_size`` > 0, every traced
 // frame encountered (leaf-first) is appended to the buffer as the
@@ -224,7 +255,7 @@ static int whereInPythonImpl(std::string& filename, int& lineno, int& bytei,
     PyPtr<PyCodeObject> code =
         PyFrame_GetCode(static_cast<PyFrameObject*>(frame));
     PyPtr<> co_filename =
-        PyUnicode_AsASCIIString(static_cast<PyCodeObject*>(code)->co_filename);
+        encodeFilename(static_cast<PyCodeObject*>(code)->co_filename);
 
     if (!(static_cast<PyObject*>(co_filename))) {
       if (stack_bytes_written != nullptr) *stack_bytes_written = written;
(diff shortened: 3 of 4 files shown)


emeryberger:
Added a `directory-only` layout to the end-to-end test (90cabf7a), covering the reporter's exact framing — "a module in its import tree which is **at a path** with non-ascii characters in it," where the umlaut is not in the filename at all.

The three layouts are now explicit:

| id | path | non-ASCII component |
|---|---|---|
| `leaf-only` | `program/überschüsse.py` | filename |
| `directory-only` | `optionale_verlängerung/pakete/workload.py` | interior directory only — module *and* its immediate parent are ASCII |
| `both` | `prögramm/überschüsse.py` | both |

The native code encoded the whole path, so a single non-ASCII character anywhere in it was enough. Verified `directory-only` against the parent commit's native code:

```
'ascii' codec can't encode character '\xe4' in position 130: ordinal not in range(128)
  File ".../optionale_verlängerung/pakete/workload.py", line 6, in f
```

That `\xe4` is the `ä` from the directory, on an all-ASCII `workload.py`. All four tests pass with the fix.

